PR #7, opened 2017-01-19
Prep next release

rm-hull:


luma/core/emulator.py
```diff
@@ -198,8 +198,11 @@ class transformer(object):
     """
     def __init__(self, pygame, width, height, scale):
         self._pygame = pygame
+        self._input_size = (width, height)
         self._output_size = (width * scale, height * scale)
-        self.scale = scale
+        self._scale = scale
+        self._led_on = self._pygame.image.load(os.path.join(os.path.dirname(__file__), "images", "led_on.png"))
```
thijstriemstra: I know we don't have, or need to, introduce a fixed width for a line, e.g. 80, but these statements, _led_on and _led_off, are more readable when kept under 80 imo

rm-hull: sure - lets implement a flake8 policy for this and enforce it though? - as another issue / PR though, as I think there will be some widescale reformatting needed

thijstriemstra: yea let's do that in the post-release develop branch

PR #16, opened 2017-01-27
GIF Anim - bug fix & optimize palette on save

rm-hull:


luma/core/emulator.py
```diff
@@ -114,15 +114,16 @@ def display(self, image):
...
-                                 append_images=self._images[1:],
-                                 format="GIF")
-
-        logger.debug("Wrote {0} frames to file: {1} ({2} bytes)".format(
-            len(self._images), self._filename, os.stat(self._filename).st_size))
+        if len(self._images) > 0:
+            logger.debug("Please wait... building animated GIF")
+            with open(self._filename, "w+b") as fp:
+                self._images[0].save(fp, save_all=True, loop=self._loop,
+                                     duration=int(self._duration * 1000),
```
thijstriemstra: for clarity, I'd move that calculation to a separate var and just do `duration=duration,`.

luma/core/emulator.py
```diff
@@ -114,15 +114,16 @@ def display(self, image):
...
+                                     append_images=self._images[1:],
+                                     optimize=True, format="GIF")
+
+            logger.debug("Wrote {0} frames to file: {1} ({2} bytes)".format(
+                len(self._images), self._filename, os.stat(self._filename).st_size))
```
thijstriemstra: same for the `os.stat` call, let's move it to a separate var.

PR #24, opened 2017-02-22
Make pytest-runner a conditional requirement

7754359337:
Hi, I am trying to embed luma into my project which uses Buildroot Embedded Linux. Pytest is not provided at the time of installation and so it fails.

I feel like this is a good compromise and it is also what [other packages seem to be doing](https://github.com/thombashi/DataProperty/issues/19).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ setup(
     namespace_packages=["luma"],
     packages=["luma.core", "luma.core.legacy"],
     install_requires=["pillow>=4.0.0", "smbus2", "spidev", "RPi.GPIO"],
-    setup_requires=[] + pytest_runner
+    setup_requires=[] + pytest_runner,
     tests_require=["mock", "pytest", "pytest-cov", "python-coveralls"],
     zip_safe=False,
     classifiers=[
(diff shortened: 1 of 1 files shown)


7754359337:
I'm sorry, I think that the `.travis.yml` would also need modifying. I think if you added `python setup.py test` into the scripts section it should fix this.

rm-hull:
@7754359337 Thanks for the PR .. what you're suggesting seems entirely reasonable, but would it be better to add the change to the [tox.ini](https://github.com/rm-hull/luma.core/blob/master/tox.ini#L10) file instead? Can you make the changes on you fork, and it should just rebuild on travis

I see there's a PR for the luma.oled package as well .. dont suppose you fancy doing PRs for the luma.led_matrix and luma.lcd as well to keep them all aligned?  :-]

setup.py
```diff
@@ -23,7 +27,7 @@
     namespace_packages=["luma"],
     packages=["luma.core", "luma.core.legacy"],
     install_requires=["pillow>=4.0.0", "smbus2", "spidev", "RPi.GPIO"],
-    setup_requires=["pytest-runner"],
+    setup_requires=[] + pytest_runner
```
rm-hull: Requires a comma at the end of the line

PR #27, opened 2017-02-26
Add spritesheet and framerate_regulator functionality

rm-hull:


luma/core/sprite_system.py
```diff
@@ -0,0 +1,180 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2014-17 Richard Hull and contributors
+# See LICENSE.rst for details.
+
+import time
+from PIL import Image
+
```
thijstriemstra: can this module get a docstring? I'd also use 2017 only in the copyright string to remind it's only available recently.

luma/core/sprite_system.py
```diff
@@ -0,0 +1,180 @@
...
+                setattr(self, a, dict_wrapper(b) if isinstance(b, dict) else b)
+
+
+class spritesheet(object):
+    """
+    A sprite sheet is a series of images (usually animation frames) combined
+    into a larger image. A dictionary is usually spread into the object
+    constructor with the following attributes:
+
+        * ``image``: a string representing the path to a sprite map image.
```
thijstriemstra: i'd prefer to use the Sphinx/rst way of documenting parameters and attributes, what about you?

PR #28, opened 2017-02-26
Remove unnecessary dependencies for test builds

thijstriemstra:


thijstriemstra:
`i2c-tools` is still required to trigger the i2c permission error.

PR #30, opened 2017-02-27
Add full-frame and diff-to-previous framebuffer implementations

rm-hull:
This is an enabler, and other functionality to follow in specific
device implementations, notably SSD1331, SSD1322, SSD1325 - which
should see improved render performance with the ‘diff_to_previous’
frame buffer

### DIFF
--- a/luma/core/framebuffer.py
+++ b/luma/core/framebuffer.py
@@ -1,0 +1,85 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2014-17 Richard Hull and contributors
+# See LICENSE.rst for details.
+
+"""
+Different implementation strategies for framebuffering
+"""
+
+from PIL import Image, ImageChops
+
+
+class diff_to_previous(object):
+    """
+    Compare the current frame to the previous frame and tries to calculate the
+    differences: this will either be ``None`` for a perfect match or some
+    bounding box describing the areas that are different, upto the size of the
+    entire image.
+
+    The image data for the difference is then be passed to a device for
+    rendering just those small changes. This can be very quick for small screen
+    updates, but suffers from variable render times, depending on the changes
+    applied. The :py:class`luma.core.sprite_system.framerate_regulator` may be
+    used to counteract this behviour however.
+
+    :param device: the target device, used to determine the initial 'previous'
+        image.
+    :type device: luma.core.device.device
+    """
+    def __init__(self, device):
+        self.image = Image.new(device.mode, device.size, "white")
+
+    def update(self, image):
+        """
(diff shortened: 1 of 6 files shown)


luma/core/framebuffer.py
```diff
@@ -0,0 +1,85 @@
...
+class full_frame(object):
+    """
+    Always renders the full frame every time. This is slower than
+    :py:class:`diff_to_previous` as there are generally more
+    pixels to update on every render, but it has a more consistent render time.
+    Not all display drivers may be able to use the differencing framebuffer, so
+    this is provided as a drop-in replacement.
+
+    :param device: the target device, used to determine the bounding box.
+    :type device: luma.core.device.device
```
thijstriemstra: use 
```
:py:class:`luma.core.device.device`
```

luma/core/framebuffer.py
```diff
@@ -0,0 +1,85 @@
...
+        self.bbox = ImageChops.difference(self.image, image).getbbox()
+        if self.bbox is not None:
+            self.image = image.copy()
+
+    def getdata(self):
+        """
+        A sequence of pixel data relating to the changes that occurred
+        since the last time :py:func:`update` was last called.
+
+        :returns: A sequence of pixels
```
thijstriemstra: `:rtype: list`

luma/core/framebuffer.py
rm-hull: Is this necessary ... it renders it as a link anyway:
screen shot 2017-02-27 at 00 42 11

luma/core/framebuffer.py
rm-hull: Strictly speaking it is not an image, but an iterable of (a tuple of RGB bands or a single intensity value):

```
 $ python
Python 2.7.13 (default, Dec 19 2016, 01:44:42) 
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from PIL import Image
>>> x = Image.new("1", (40,40))
>>> x
<PIL.Image.Image image mode=1 size=40x40 at 0x10167CB50>
>>> x.getdata()
<ImagingCore object at 0x101567190>
```

thijstriemstra: iterable instead of list

luma/core/framebuffer.py
thijstriemstra: ah shit didnt know that, cool

thijstriemstra: i suppose for external modules it is

PR #35, opened 2017-02-27
Api docs

rm-hull:


luma/core/serial.py
```diff
@@ -2,22 +2,31 @@
...
 
 
 class i2c(object):
     """
-    Wrap an `I2C <https://en.wikipedia.org/wiki/I%C2%B2C>`_ interface to
-    provide data and command methods.
-
-    :param bus: I2C bus instance.
+    Wrap an `I²C <https://en.wikipedia.org/wiki/I%C2%B2C>`_ (Inter-Integrated
+    Circuit) interface to provide :py:func:`data` and :py:func:`command` methods.
```
thijstriemstra: I have a suspicion it's these unicode things, e.g. I²C, is screwing up the intersphinx build.

rm-hull: Seemed to work ok when I built locally, and I have used the unicode degrees before

PR #37, opened 2017-03-06
More coverage

thijstriemstra:


tests/test_framerate_regulator.py
```diff
@@ -58,3 +63,15 @@ def test_sleep():
 
     assert regulator.called == 200
     assert after - before >= 2.0
+
+
```
thijstriemstra: the 2 tests below indicate there's an issue with calling them directly, something that should be probably fixed (in this or other PR).

tests/test_virtual.py
```diff
@@ -137,6 +142,14 @@ def draw_fn(draw, width, height):
...
 
     bbox = ImageChops.difference(reference, device.image).getbbox()
     assert bbox is None
+
+
+def test_terminal_println():
+    device = dummy()
+    ter = terminal(device)
+
+    ter.println('foo\rbar\bspam\teggs\n')
```
thijstriemstra: not sure what to assert here but the coverage is there.

tests/test_framerate_regulator.py
rm-hull: Agreed - let's do it as a separate PR. If count is zero or start time is None, both those methods could return None

tests/test_virtual.py
rm-hull: I think the best you can hope for here is a regression test - much like the others in this file:

* load a reference image
* do the operation
* diff the resulting image with the reference image 
* assert that the bounding box is None

In order to generate the reference image, I would just do a one-off `(device.image.save('tests/reference/terminal.py')` and visually check that the resultant image looks right.

Also, this technique can be used to check that the terminal lines up tabs, eg.

```python
ter.println("1\t32\t999\n")
ter.println("999\t1\t32\n")
```

And double newlines should have a blank line between text,

```python
ter.println("line1\n\nline2\n")
```

Lastly, check that wrapping to the device width works as expected

tests/test_dummy.py
```diff
@@ -30,7 +33,8 @@ def test_portrait():
     device = dummy(rotate=1)
 
     # Use the same drawing primitives as the demo
-    with canvas(device) as draw:
+    # Enable dither for extra test coverage
+    with canvas(device, dither=True) as draw:
```
rm-hull: Would be better to create a separate test for dither - create a `mode="1"` dummy device, and draw some coloured boxes onto a dithered canvas, and check that resultant image is dithered. Whether pillow's dither implementation works consistently/randomly may thwart this though

rm-hull: But lets do that separately.

rm-hull: See https://github.com/rm-hull/luma.core/pull/39

thijstriemstra:
Thanks for the feedback. Feel free to adjust things in this branch, I won't have time to work on it tonight.

rm-hull:
Sure thing - once again: awesome effort - chapeau!

PR #38, opened 2017-03-06
add error.UnsupportedPlatform

thijstriemstra:


thijstriemstra:
This should fix: https://github.com/rm-hull/luma.lcd/issues/28

luma/core/error.py
```diff
@@ -36,3 +36,11 @@ class DeviceDisplayModeError(Error):
     """
     Exception raised when an invalid device display mode is detected.
     """
+
+
+class UnsupportedPlatform(Error):
+    """
+    Exception raised when trying to use the library on an incompatible system.
+
+    .. versionadded:: 0.5.2
```
thijstriemstra: we're not sure if the next version will be 0.5.2 but .. let's update it (or try to remember to do that) when its release time.

PR #39, opened 2017-03-06
Add test to check canvas dither

rm-hull:


tests/test_dummy.py
```diff
@@ -35,3 +35,22 @@ def test_portrait():
 
     bbox = ImageChops.difference(reference, device.image).getbbox()
     assert bbox is None
+
+
+def test_dither():
+    reference = Image.open(
```
thijstriemstra: can you add a docstring?

PR #45, opened 2017-03-08
Handle rpi.gpio import error

thijstriemstra:


thijstriemstra:
@rm-hull unfortunately this needs to be released again.

PR #46, opened 2017-03-08
fix lib decorators

thijstriemstra:


thijstriemstra:
Also will add tests.

thijstriemstra:
@rm-hull should be fixed now, but requires a release again.

PR #48, opened 2017-03-09
Add ANSI colors to luma.core.virtual.terminal

### LINKED ISSUE #40: Add ANSI colors to luma.core.virtual.terminal
`puts()` processes a string of characters one at time, but this could be changed to be made aware of ANSI escape sequences, particularly the color codes below.

On encountering a new (valid) color code, the `self.color` and/or `self.bgcolor` could be set accordingly. Also need to consider what the reset code (`ESC[0m`) would reset the colors to.

screen shot 2017-03-06 at 21 55 35

(screen grab taken from http://ascii-table.com/ansi-escape-sequences.php)

rm-hull:
Fixes #40


luma/core/ansi_color.py
```diff
@@ -0,0 +1,49 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2017 Richard Hull and contributors
+# See LICENSE.rst for details.
+
```
thijstriemstra: can you add a docstring?

luma/core/ansi_color.py
```diff
@@ -0,0 +1,49 @@
...
+    42:  ["background_color", "green"],
+    43:  ["background_color", "yellow"],
+    44:  ["background_color", "blue"],
+    45:  ["background_color", "magenta"],
+    46:  ["background_color", "cyan"],
+    47:  ["background_color", "white"],
+}
+
+
+def parse_str(text):
```
thijstriemstra: parse_ansi_str?

luma/core/ansi_color.py
thijstriemstra: and a version added directive in that docstring.

luma/core/ansi_color.py
rm-hull: In context of `ansi_color.parse_ansi_str`, it seems a bit of a mouthful. 
In isolation `parse_ansi_str` is probably ok, but generally: what benefit/drawback does adding 'ansi' into the method name convey?

Issue #2, opened 2017-01-12
Deprecate SPI params bcm_DC, bcm_RST in favour of gpio_DC, gpio_RST


rm-hull:
See https://github.com/rm-hull/luma.core/blob/master/luma/core/serial.py#L96

*bcm* is shorthand for 'broadcom' pin assignment, and is specific to RPi implementations, whereas *gpio* is a more general term, and should be used if ever a generic GPIO library replaces RPi.GPIO

thijstriemstra:
perhaps we could use python's DeprecationWarning here

PR #49, opened 2017-03-09
Catch IO error

thijstriemstra:


thijstriemstra:
@rm-hull can you add a new release?

rm-hull:
done ==> 0.6.0 released with these changes in

thijstriemstra:
Thanks! I'm now working on #2.

PR #50, opened 2017-03-13
Deprecate spi params

### LINKED ISSUE #2: Deprecate SPI params bcm_DC, bcm_RST in favour of gpio_DC, gpio_RST

thijstriemstra:
fixes #2

### DIFF
--- a/luma/core/sprite_system.py
+++ b/luma/core/sprite_system.py
@@ -67,7 +67,8 @@ class spritesheet(object):
     Loosely based on http://www.createjs.com/docs/easeljs/classes/SpriteSheet.html
     """
     def __init__(self, image, frames, animations):
-        self.image = Image.open(image)
+        self._img = open(image, 'rb')
+        self.image = Image.open(self._img)
         self.frames = dict_wrapper(frames)
         self.animations = dict_wrapper(animations)
         # Reframe the sprite map in terms of the registration point (if set)
@@ -124,7 +125,7 @@ class spritesheet(object):
         return self.frames.count
 
     def close(self):
-        self.image.close()
+        self._img.close()
 
     def animate(self, seq_name):
         """
--- a/luma/core/sprite_system.py
+++ b/luma/core/sprite_system.py
@@ -67,8 +67,9 @@ class spritesheet(object):
     Loosely based on http://www.createjs.com/docs/easeljs/classes/SpriteSheet.html
     """
     def __init__(self, image, frames, animations):
-        self._img = open(image, 'rb')
-        self.image = Image.open(self._img)
+        with open(image, 'rb') as fp:
+            self.image = Image.open(fp)
+            self.image.load()
         self.frames = dict_wrapper(frames)
         self.animations = dict_wrapper(animations)
         # Reframe the sprite map in terms of the registration point (if set)
@@ -123,9 +124,6 @@ class spritesheet(object):
         The number of frames in the sprite sheet
         """
         return self.frames.count
-
-    def close(self):
-        self._img.close()
 
     def animate(self, seq_name):
         """
--- a/luma/core/sprite_system.py
+++ b/luma/core/sprite_system.py
@@ -10,6 +10,7 @@ Simplified sprite animation framework.
 """
 
 import time
+
 from PIL import Image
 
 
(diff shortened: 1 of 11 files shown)


thijstriemstra:
work in progress..

thijstriemstra:
Enabling warnings displays a lot of resource warnings: https://travis-ci.org/rm-hull/luma.core/jobs/210702792#L472. I'll fix those as well.

thijstriemstra:
@rm-hull the resource warnings coming from `test_spritesheet` could be considered bugs because the luma code doesn't close the image for you. But I'm not sure where to close the image in the `spritesheet` class..

```
WW0 in tests/test_spritesheet.py:42 the following warning was recorded:
 .tox/py35/lib/python3.5/site-packages/PIL/Image.py:1028: ResourceWarning: unclosed file <_io.BufferedReader name='tests/reference/runner.png'>
  self.load()
```

tests/test_serial.py
```diff
@@ -138,14 +138,42 @@ def verify_spi_init(port, device, bus_speed_hz=8000000, dc=24, rst=25):
...
     ])
 
 
+def test_spi_init_params_deprecated():
+    port = 5
+    device = 2
+    bus_speed = 16000000
+    dc = 80
+    rst = 90
+    msg1 = 'bcm_DC argument is deprecated in favor of gpio_DC and will be removed in 1.0.0'
```
thijstriemstra: @rm-hull this is an arbitrary version number, what would be a good version to drop this?

rm-hull: 1.0.0 is probably the right version ... as long as we dont forget!

rm-hull:
Think you're right about the spritesheet - maybe adding a `__del__()` destructor method to the spritesheet class that calls `self.image.close()` ?

thijstriemstra:
@rm-hull spritesheet probably needs a `close` method. This can be called manually and if spritesheet would support context manager usage, `close` could be called there.

tests/test_serial.py
thijstriemstra: I'll open a ticket for it.

luma/core/sprite_system.py
```diff
@@ -122,6 +124,9 @@ def __len__(self):
         """
         return self.frames.count
 
+    def close(self):
+        self._img.close()
+
```
thijstriemstra: unfortunately self.image.close() didn't work so added a new private var that can be closed.

thijstriemstra:
Should be ready for review.

rm-hull:
Just reading the docs for https://pillow.readthedocs.io/en/4.0.x/reference/Image.html#PIL.Image.Image.load

This implies as soon as an image is accessed, `load` is called and the underlying file pointer is closed. Are we unnecessarily complicating things? Are the warnings actually just false alarms?

thijstriemstra:
Pillow seems buggy with closing them. Pillow should also support the `with` keyword but it's not working. Also see http://stackoverflow.com/a/38542329

luma/core/sprite_system.py
```diff
@@ -66,7 +67,8 @@ class spritesheet(object):
     Loosely based on http://www.createjs.com/docs/easeljs/classes/SpriteSheet.html
     """
     def __init__(self, image, frames, animations):
-        self.image = Image.open(image)
+        self._img = open(image, 'rb')
+        self.image = Image.open(self._img)
```
rm-hull: this works (tested), and doesn't rely on the consumer to call close:

```python
    def __init__(self, image, frames, animations):
        with open(image, "rb") as fp:
            self.image = Image.open(fp)
            self.image.load()

        ...
```

presumably it wont trigger the warning

thijstriemstra: thanks that worked. updated PR.

rm-hull:
Can you squash & merge when everything goes green pls ?

thijstriemstra:
I guess the examples now need to updated, working on PR.

PR #53, opened 2017-03-14
Move GPIO.setmode() to point when referenced

rm-hull:


luma/core/lib.py
```diff
@@ -20,8 +20,9 @@ def __rpi_gpio__(self):
     # a Raspberry Pi... this is imported here so we can swap out the
     # implementation for a mock
     try:
-        import RPi.GPIO
-        return RPi.GPIO
+        import RPi.GPIO as GPIO
+        GPIO.setmode(GPIO.BCM)
```
thijstriemstra: hmm, what the rationale behind this change?

rm-hull: Somewhat related to the ticket you just closed (#2): If you _dont_ supply a gpio= override, then the behaviour is unchanged from present - it sets up the RPi.GPIO with BCM mode.

For those times when gpio= is provided, the chances are the reason is that it needs to integrate with an existing codebase that uses a RPi.GPIO which has already been setup and initialized: we shouldn't assume BCM is the right mode for all applications & should let the caller choose.

Plus ... I have an alternative [OPi.GPIO](https://github.com/rm-hull/OPi.GPIO) implementation under development where it make more sense to use the SUNXI pin mappings with the Orange Pi Zero... [the implementation does support translating BCM to native pin mappings though]

Same goes for the CHIP_IO, which I suspect _probably_ just ignores the `GPIO.setmode(GPIO.BCM)` (haven't checked)

thijstriemstra: I guess I'm concerned about the amount of times this code can be called.. It could cause a lot of setmode calls whereas these imports are less harmless.

thijstriemstra: you say:

> we shouldn't assume BCM is the right mode for all applications & should let the caller choose.

then this change contradicts that doesn't it? it hardcodes BCM.

rm-hull: For the default case, it does initialize RPi.GPIO with the BCM mode when `gpio=None` is supplied.

But when supplying a specific GPIO module to the spi class, we shouldn't enforce it to be BCM.... Say I have an existing application, which I've coded up to already use RPi.GPIO with BOARD mode, in order to make use of it I would need to pass in `gpio=RPi.GPIO`, and then supply `gpio_DC=<board pin>`, `gpio_RST=<board pin>`

Or in the case of orange Pi, I would initialize like:

```python
import OPi.GPIO as GPIO
from luma.core.serial import spi

GPIO.setmode(GPIO.SUNXI)

spi = serial(gpio=GPIO, gpio_DC="PA01", gpio_RST="PG06", ...)
```

Right now, the above code would not be able to work.

The `setmode` method has just moved, and would get initialized once per newly created `spi` class, or in the case of when `gpio=GPIO` is supplied it is at the end users discretion to set the mode.

PR #62, opened 2017-03-22
Add utilities from luma.examples

### LINKED ISSUE #58: add argparser utils from luma.examples
Being able to reuse the demo_opts, argparser code, in luma.examples would be beneficial for sub-projects. You would be able to use the parser code to create a command-line script that handles all compatible luma libraries.

I'm not sure if it belongs in core or a new luma.commandline-something llibrary. Thoughts?

thijstriemstra:
fixes #58


thijstriemstra:
work in progress..

luma/core/serial.py
```diff
@@ -225,7 +225,7 @@ class noop(object):
     def __getattr__(self, attr):
         return self.__noop
 
-    def __setattr__(self, attr, val):
+    def __setattr__(self, attr, val):  # pragma: no cover
         pass
 
     def __noop(*args, **kwargs):
```
thijstriemstra: @rm-hull is there a `self` missing here?

thijstriemstra:
I'm gonna add a few more tests but it's basically ready for review @rm-hull

luma/core/serial.py
rm-hull: Ooh, looks like it...

tests/test_framerate_regulator.py
```diff
@@ -66,11 +64,11 @@ def test_sleep():
...
-        regulator.effective_FPS()
+    assert regulator.effective_FPS() == 0
 
 
 def test_average_transit_time():
     regulator = framerate_regulator(fps=30)
-    with pytest.raises(ZeroDivisionError):
-        regulator.average_transit_time()
+    with regulator:
+        pass
```
rm-hull: Should it also check the functions return value before the with block?

PR #63, opened 2017-03-22
Bitbang SPI

### LINKED ISSUE #15: Support bit-bang SPI
Add an implementation that uses (any) GPIO pins to simulate the SPI protocol in software. It is likely to be slow, but useful for those devices which are hardwired, but not to the SPI pins (WTF?) eg: https://www.amazon.co.uk/SunFounder-Raspberry-PCD8544-Backlight-Compatible/dp/B01GPV7M1W/ref=sr_1_1?ie=UTF8&qid=1485198921&sr=8-1&keywords=pcd8544

See also: https://github.com/adafruit/Adafruit_Python_GPIO/blob/master/Adafruit_GPIO/SPI.py#L150

rm-hull:
This is an enabler for APA102C RGB LEDs such as [BLINKT!](https://shop.pimoroni.com/products/blinkt) and some non-standard PCD8544 breakout boards

Fixes #15


luma/core/serial.py
```diff
@@ -111,12 +112,91 @@ def cleanup(self):
...
-        self._bus.close()
+        if self._managed:
+            self._bus.close()
 
 
-@lib.spidev
 @lib.rpi_gpio
-class spi(object):
+class bitbang(object):
+
```
thijstriemstra: class docstring + version added

tests/test_bitbang.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# Copyright (c) 2017 Richard Hull and contributors
+# See LICENSE.rst for details.
+
+"""
+Tests for the :py:class:`luma.core.serial.bitbang` class.
+"""
+
+try:
```
thijstriemstra: theres `helpers` you can import this from

PR #75, opened 2017-04-12
don't rely on i2c-tools in tests

### LINKED ISSUE #61: Use pyfakefs to simulate /dev/i2c-1 device in tests
* test_i2c_init_device_not_found
* test_i2c_init_device_permission_error

thijstriemstra:
fixes #61

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: true
+sudo: false
 language: python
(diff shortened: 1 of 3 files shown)


thijstriemstra:
we don't even need pyfakefs, it can be done with mock instead. I'll simplify the tests.

Issue #41, opened 2017-03-06
Add 4- & 8-bit parallel interfaces


rm-hull:
Somewhat related to https://github.com/rm-hull/luma.lcd/issues/19

Use RPi.GPIO to bit-bang out 4 or 8 bus lines according to the HD44780 datasheet spec. The interface should have the same "shape" as the existing [luma.core.interface.serial](https://github.com/rm-hull/luma.core/blob/master/luma/core/interface/serial.py) interfaces - i.e. implement `data()` and `command()` methods.

Also: ~~Move `luma.core.serial` to `luma.core.interface.serial` and add this new functionality to `luma.core.interface.parallel`. Various breaking changes to device implementations to rename `serial_interface=` constructor arg to `interface=`.~~ (done with #78)

<img width="616" alt="screen shot 2017-03-06 at 22 12 28" src="https://cloud.githubusercontent.com/assets/1915543/23632515/1767fcca-02ba-11e7-8d74-a9f636e806ba.png">

(screen grab from https://cdn-shop.adafruit.com/datasheets/HD44780.pdf)

thijstriemstra:
> Move luma.core.serial to luma.core.interface.serial and add this new functionality to luma.core.interface.parallel. Various breaking changes to device implementations to rename serial_interface= constructor arg to interface=.

Let's add a proper DeprecationWarning when importing from the old location. The warning should also include the version nr when it'll be removed.

PR #78, opened 2017-04-16
add word wrap capability to luma.core.virtual.terminal

### LINKED ISSUE #42: Add word-wrap capability to luma.core.virtual.terminal
Currently, the terminal class will check that text to be rendered using proportional fonts wont overflow over the right hand edge. If there is not enough width to display the next character, it will insert a newline. For example:

image

Add a constructor arg setting so that wrapping works at the word level, eg, the above string would render as

```
The quick brown fox
jumps over the lazy
dog
```

thijstriemstra:
* fixes #42 
* adds `luma.core.interface` package (#41)
* move `luma.core.serial` to `luma.core.interface.serial`
* deprecate `luma.core.serial` module


luma/core/virtual.py
```diff
@@ -227,8 +229,14 @@ def println(self, text=""):
         Prints the supplied text to the device, scrolling where necessary.
         The text is always followed by a newline.
         """
-        self.puts(text)
-        self.newline()
+        if self.word_wrap:
+            text = wrap(text, width=self.width)
```
rm-hull: Does wrap work with ANSI color codes? Would be worth adding a test to check.

doc/api-documentation.rst
```diff
@@ -37,6 +37,13 @@ API Documentation
     :undoc-members:
     :show-inheritance:
 
+:mod:`luma.core.interface.serial`
+"""""""""""""""""""""""
```
rm-hull: Think the underline needs to be the same length as the title else doc generation will complain. we should probably add docgen to the toxic was task if possible

luma/core/virtual.py
thijstriemstra: I've added a test but I'm not sure if it's correct, probably not and we need to ignore out these ansi color definitions?

ansi_colors_wrapped

thijstriemstra: it's def not correct, 'hello world' line is missing. hmm any ideas @rm-hull?

rm-hull: Hello world is missing because it has scrolled off the top. Clearly the word wrap is counting the ANSI colour code characters rather than interpreting them as zero width.

Wonder whether we need a home grown text wrapper that does take ANSI color codes into account?

Maybe just leave as-is for now.

Interestingly, it has exposed another bug (which was there before your commits) when scrolling when the background colour is not black - see the yellow bar at the bottom.

thijstriemstra: I want to look into the issue a bit more before merging, thanks for the feedback.

thijstriemstra: fixed it!

ansi_colors_wrapped

@rm-hull also used version added `0.9.0`, I assume that'll be the next version?

thijstriemstra: > Interestingly, it has exposed another bug (which was there before your commits) when scrolling when the background colour is not black - see the yellow bar at the bottom.

I've opened #79 for that issue.

PR #86, opened 2017-06-19
Adding optional background image parameter to canvas

TheRaspyDev:
I wanted to have an easy way to write text on a background image.
That's why I added an optional background parameter when creating the canvas.

### DIFF
--- a/luma/core/render.py
+++ b/luma/core/render.py
@@ -17,9 +17,12 @@ class canvas(object):
     space into a 1-bit monochrome image where dithering is employed to
     differentiate colors at the expense of resolution.
     """
-    def __init__(self, device, dither=False):
+    def __init__(self, device, background=None, dither=False):
         self.draw = None
-        self.image = Image.new("RGB" if dither else device.mode, device.size)
+        if background is None:
+            self.image = Image.new("RGB" if dither else device.mode, device.size)
+        else:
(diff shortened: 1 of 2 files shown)


luma/core/render.py
```diff
@@ -17,9 +17,12 @@ class canvas(object):
...
     differentiate colors at the expense of resolution.
     """
-    def __init__(self, device, dither=False):
+    def __init__(self, device, background=None, dither=False):
         self.draw = None
-        self.image = Image.new("RGB" if dither else device.mode, device.size)
+        if background is None:
+            self.image = Image.new("RGB" if dither else device.mode, device.size)
+        else:
+            self.image = background.copy()
```
rm-hull: I think it should `assert(background.size == device.size)` here just to make sure that the incoming canvas size matches what the device is capable of.

TheRaspyDev:
Thanks for the fast review.
Added the assertion, the contribution and the parameter description.
I don't know how to write the regression test so please do it for me :)

rm-hull:
@TheRaspyDev - v0.9.2 has been released with your code change. 
FYI, have a look at commit d7ed02536d6b4d36d587ad487bda2661f8bd2349 for the test I added.
Once again, thanks for your contribution!

PR #91, opened 2017-07-04
Pytest changed the deprecated_call() internals

rm-hull:


rm-hull:
Don't know - but we'll only ever run with the latest pytest?

PR #110, opened 2017-11-22
Declare __version__ number in library

thijstriemstra:
I'm using pyqt (5) and this is the only way I can grab the version nr (from the lib) unless I also include the txt file with version nr (for each luma lib) and read it independently. 

Goal:

```python
from luma.core import __version__
print(__version__)  # '1.2.3'
```


thijstriemstra:
Going with idea from https://packaging.python.org/guides/single-sourcing-package-version/ here.

thijstriemstra:
Looks like python 2.7 is only one complaining now.

PR #114, opened 2017-11-26
use extras_require for linux dependencies

thijstriemstra:
to avoid invalid wheels with setuptools<36.2.0 (h/t @benoit-pierre)


thijstriemstra:
:+1:

PR #120, opened 2018-02-05
7 Segment Font

jonathanrjpereira:
A font for Dot Matrix Display based on [7 Segment Font](https://www.dafont.com/digital-7.font)
Each character consumes an area of 3 columns X 7 rows.
Supports only characters from 0x41 - 0x5A & 0x30 - 0x39 inclusive.
Special Characters are not supported. Uppercase & Lowercase letters have the same bit patterns.

I created the font with the intention of displaying static alpha-numeric text.
8 different characters can be statically displayed on a 4in1 cascaded matrix display.  i.e.: 2 characters in an 8x8 matrix.


rm-hull:
Version 1.5.0 released with these changes

Issue #124, opened 2018-02-20
Higher-level abstractions for setting SPI transfer_size


drspangle:
This is a feature request, not a bug report.

Currently, the command line parsers and device configuration settings for the MAX7219 devices in the luma library allow the SPI bus speed (i.e. clock speed) and CS active-high to be set, but there's no way to adjust the maximum `transfer_size` parameter, which is exposed by the `spi` class constructor (the default is 4096). I think this would be very useful for me to debug my glitchy MAX7219-based displays; the raspberry pi supports 4096, but I think the cascaded MAX7219s in my display would be happier with smaller words. I don't want to have to manually override the default parameter in the core library, it would be great if this could be supported from the command line argument parser.

drspangle:
@rm-hull Would I be correct to assume that you're going to implement this feature at some point in the near future? If there's anything I can do to help, please let me know.

thijstriemstra:
Maybe you can try it, it's exactly the same as https://github.com/rm-hull/luma.core/commit/5c688227fc26d0373a92e778813dc60fb514aeba but then for transfer_size @drspangle

drspangle:
@thijstriemstra Fair enough, I created a new PR #125 based on that commit you linked.

rm-hull:
I was going to, but you beat me to it

rm-hull:
Will create a new release tomorrow for this

PR #125, opened 2018-02-21
Fix for Issue #124

drspangle:
Added command line argument that allows the maximum transfer unit for SPI to be adjusted.

Included the new configurable `spi_transfer_size` field within the `test_spi_opts` class for the command line argument parser module.

### DIFF
--- a/luma/core/cmdline.py
+++ b/luma/core/cmdline.py
@@ -155,6 +155,7 @@ class make_serial(object):
                    device=self.opts.spi_device,
                    bus_speed_hz=self.opts.spi_bus_speed,
                    cs_high=self.opts.spi_cs_high,
+                   transfer_size=self.opts.spi_transfer_size,
                    gpio_DC=self.opts.gpio_data_command,
                    gpio_RST=self.opts.gpio_reset,
                    gpio=self.gpio or GPIO)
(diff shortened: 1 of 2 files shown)


drspangle:
Just a quick note regarding the CI, it looks like the only tests that failed were because of a misconfigured CI environment. From what I can tell, all the unit tests passed - this should trivally pass because the unit test class was only modified to include the new command line argument field in 60c3d7d, and there aren't any new tests to check the functionality (since that would be impossible to test without a physical device).

thijstriemstra:
@drspangle it's actually a pep8 style issue, see https://travis-ci.org/rm-hull/luma.core/jobs/344144854#L558

drspangle:
Weird, I must have missed that. I'll fix it.

drspangle:
@thijstriemstra What do you mean, about how the CI environment is setup? Is there some way to test the interaction with SPI, or did you mean something else?

rm-hull:
Published 1.6.0 to PiPy with this change in it.

thijstriemstra:
> Published 1.6.0 to PiPy with this change in it.

Cool you did the git squash (or some sort of commit msg amend) there as well to fix the ticket reference.

rm-hull:
Yeah... There should be some guard in github to prevent someone from trying to merge PRs on a mobile phone when drunk

PR #128, opened 2018-03-11
Optional "persist on exit"

hansliss:
Added a "persist" member to capabilities, to keep track of whether to leave the display on upon exit or clear it.
It will have to be set manually on the device variable:

```python
serial = spi(port=0, device=0, gpio=noop())
device = max7219(serial, cascaded=1)
device.persist = True
seg = sevensegment(device)
seg.text = "Hello Pi"
```

### DIFF
--- a/luma/core/device.py
+++ b/luma/core/device.py
@@ -78,7 +78,7 @@ class device(mixin.capabilities):
         helps prolong the life of the device), clear the screen and close
         resources associated with the underlying serial interface.
 
-        If self.persist is True, the device will not be switched off.
+        If :py:attr:`persist` is True, the device will not be switched off.
 
         This is a managed function, which is called when the python processs
         is being shutdown, so shouldn't usually need be called directly in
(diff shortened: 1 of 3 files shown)


tests/test_persist.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# Copyright (c) 2017-18 Richard Hull and contributors
```
thijstriemstra: minor, but this should be 2018 only.

luma/core/device.py
```diff
@@ -78,12 +78,15 @@ def cleanup(self):
         helps prolong the life of the device), clear the screen and close
         resources associated with the underlying serial interface.
 
+        If self.persist is True, the device will not be switched off.
```
thijstriemstra: i think this should be

```
:py:attr:`persist`    # or :py:var:`persist`
```

rm-hull:
Will push out a release tomorrow.

thijstriemstra:
@rm-hull can you put out a new release for this?

rm-hull:
Apologies for the delay - 1.7.0 released with this change in it

PR #140, opened 2018-05-29
support for python 3.7

thijstriemstra:
Some tests fail using Python 3.7.0b4 (final planned beta, released on 2018-05-02) on Ubuntu 16:

```
4 failed, 130 passed, 3 skipped in 4.21 seconds
```

Looks like travis is running an alpha version (Python 3.7.0a4+) and they're working on supporting the latest: https://github.com/travis-ci/travis-ci/issues/9069

Fortunately only `test_accented_charset` [fails](https://travis-ci.org/rm-hull/luma.core/jobs/385340494#L841) on that version:

```
1 failed, 133 passed, 3 skipped in 6.14 seconds
```


thijstriemstra:
@rm-hull it's failing on the generated image: https://travis-ci.org/rm-hull/luma.core/jobs/385222683 any idea why it's failing? I've always been puzzled by this commit and have the feeling it's related: https://github.com/rm-hull/luma.core/commit/1db9bff246e81a10ae56adbef9840b10bb3a15be

rm-hull:
Its something to do with rendering at the pixel level, if you look at the ä in the rich view (https://github.com/rm-hull/luma.core/commit/1db9bff246e81a10ae56adbef9840b10bb3a15be?short_path=fdd62a5#diff-fdd62a51198928d32c0adbf973b6caac)

screenshot from 2018-05-29 17-48-51

The swipe view is quite useful here... They are rendered ever so slightly differently. I can only imagine that the Pillow version has changed again, and that the rendering has maybe reverted back to how it used to be.

thijstriemstra:
I'll create a script that can be used to export the images and add it in `contrib` or `bin`.

thijstriemstra:
py3.7 final is released, let's see if travis updated it.

thijstriemstra:
Still an older version on travis but the tests now succeed on py3.7. ping @rm-hull

PR #146, opened 2018-07-27
Namespace packaging fix

Gadgetoid:
Tweak of #145 based on extensive investigation described here: https://github.com/rm-hull/luma.oled/pull/215

The `declare_namespace(__name___)` addition to `__init__.py` avoids a quirk of `setup.py/setuptools` where it will complain about namespace packages that do not include this directive, despite the `pkg_resources` approach being discouraged.

My best evidence for the reliability of this pattern is that Google use it: https://github.com/google/protobuf/blob/master/python/google/__init__.py

https://github.com/google/protobuf/blob/bdcbcabe5e123067ad2474994df343d0458b63ea/python/setup.py#L252-L257

### DIFF
--- a/luma/__init__.py
+++ b/luma/__init__.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright (c) 2014-18 Richard Hull and contributors
+# Copyright (c) 2017-18 Richard Hull and contributors
 # See LICENSE.rst for details.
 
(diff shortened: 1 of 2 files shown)


thijstriemstra:
thanks @Gadgetoid, can you add some comments in `__init__.py` that describe what's happening (and perhaps some ticket references)?

PR #147, opened 2018-08-02
Add support for smbus2 i2c_rdwr to avoid unnecessary 32byte chunking

Gadgetoid:
As discussed here; https://github.com/rm-hull/luma.oled/issues/210

In summary, the 32byte limitation is an SMBus limitation, not an i2c limitation. Since most i2c displays are closer to i2c than SMBus they're quite happy to receive a continuous buffer of N bytes, giving a slight performance increase over the chunked approach.

This change favours the use of `smbus2`'s `i2c_rdwr` method, which allows the entire display buffer or display page to be sent in a single transaction. It will use `i2c_rdwr` when in managed mode, although it will not do so if `smbus2` is supplied as the `bus` argument.

I'm not quite sure about how I've implemented this, since it presently assumes that *all* displays using this transport support continuous buffer writes of >32bytes. I generally believe this to be the case, but if there's a pattern to make this enhancement opt-in on a per device basis then I think that would be a safer approach.

Performance increases will vary across devices, but this will have the most impact where the device buffer is sent as a single transaction and not paged. Additionally, since it uses fewer transactions and returns to Python less, it should take better advantage of faster i2c baudrates. Generally increases will be *slight* (note the small increase in my sh1106 tests below).

IE: 128x128 sh1106 driven at 1MHz on Raspberry Pi (set in `/boot/config.txt`)

```text
dtparam=i2c_baudrate=1000000
```

(Note some SMBus devices will only work up to 400kHz)

Before this tweak:

```text
pi@raspberrypi:~/Development/luma.examples/examples $ python3 perfloop.py --width 128 --height 128 --display sh1106
Testing display rendering performance
Press Ctrl-C to abort test

Version: luma.oled 2.4.1 (luma.core 1.7.2)
Display: sh1106
Interface: i2c
Dimensions: 128 x 128
------------------------------------------------------------
#### iter =    217: render time = 52.09 ms, frame rate = 19.19 FPS
```

After this tweak:

```text
pi@raspberrypi:~/Development/luma.examples/examples $ python3 perfloop.py --width 128 --height 128 --display sh1106
Testing display rendering performance
Press Ctrl-C to abort test

Version: luma.oled 2.4.1 (luma.core 1.7.2)
Display: sh1106
Interface: i2c
Dimensions: 128 x 128
------------------------------------------------------------
#### iter =    310: render time = 47.13 ms, frame rate = 21.21 FPS
```

Okay it's not groundbreaking (since the sh1106 is paged), but it's an improvement!

Tested with: 

* sh1106, 128x128 on Raspberry Pi ✔️ 
* ssd1306, 128x64 on Raspberry Pi ⌛️


luma/core/interface/serial.py
```diff
@@ -55,6 +55,7 @@ def __init__(self, bus=None, port=1, address=0x3C):
 
         try:
             self._managed = bus is None
+            self._i2c_msg_write = smbus2.i2c_msg.write if bus is None else None
```
rm-hull: Wonder if lines 57-59 would be better re-written as:

```python
if bus is None: 
    self._managed = True
    self._i2c_msg_write = smbus2.i2c_msg.write
    self._bus = smbus2.SMBus(port)
else:
    self._managed = False
    self._i2c_msg_write = None
    self._bus = bus

```

thijstriemstra:
ping @Gadgetoid, any news?

thijstriemstra:
@rm-hull can you adjust the coverage treshold so it doesn't report broken builds for small changes? Maybe something like 1 or 1.5% would be better?

rm-hull:
v1.9.0 released to pypi with this improvement incorporated

rm-hull:
See also: #162

PR #149, opened 2018-09-18
Mutable string now works over unicode (for both py2/3)

rm-hull:
Goes some way to help fix test failures in https://github.com/rm-hull/luma.led_matrix/pull/158


thijstriemstra:
nice!

PR #153, opened 2018-12-22
split API docs into separate pages

### LINKED ISSUE #139: Split api docs into multiple pages
The page is becoming huge: https://luma-core.readthedocs.io/en/latest/api-documentation.html

thijstriemstra:
fixes #139

### DIFF
--- a/doc/ansi_color.rst
+++ b/doc/ansi_color.rst
@@ -1,0 +1,6 @@
+:mod:`luma.core.ansi_color`
+"""""""""""""""""""""""""""
+.. automodule:: luma.core.ansi_color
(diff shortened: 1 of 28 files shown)


thijstriemstra:
@rm-hull can you push a new release for this change because API links seem to be broken now, e.g. check the core api links on https://luma-led-matrix.readthedocs.io/en/latest/python-usage.html. Hopefully that will fix it..

rm-hull:
Read-the-docs pages get built on push to master, so I don’t think a release will help. Do we just need to update the links to the changed pages in the led matrix project instead?

PR #154, opened 2018-12-30
improve cmdline help output

thijstriemstra:
fixes https://github.com/rm-hull/luma.examples/issues/96


thijstriemstra:
In order for this to show up in the examples, a new core release is needed. Can you do this @rm-hull ?

rm-hull:
v1.8.3 released

thijstriemstra:
Thanks!

PR #159, opened 2019-04-09
FTDI wrapper

rm-hull:
Using an FT232H board to bridge between USB and SPI/I2C to allow luma display drivers to be used on a regular PC where GPIO pins are not generally exposed. 

These boards can be picked up from ebay or aliexpress for a few pounds:

image

# TODO

* [x] SPI wrapper
* [x] I2C wrapper
* [x] GPIO wrapper
* [x] `fdti_i2c` serial interface
* [x] `fdti_spi` serial interface
* [x] Regression tests
* [ ] Documentation
* [x] Cmdline helpers


thijstriemstra:
very interesting!

rm-hull:
> bring on those FPS!

Initial indications are that it is very fast on my linux workstation, at least for SPI. 

Will be interesting to see if it work cross platform - I can try OSX but haven't got Windows 10... of course this will add another dimension to try to support it &lt;groan&gt;

thijstriemstra:
> very fast 

numbers or it didn't happen ;)

rm-hull:
Unfortunately I'm away now until Thursday, and i forgot to bring the board with me. I did run the perfloop but I honestly can't remember what the value was... sorry to leave it as a cliffhanger! :laughing: 

If it's any consolation, here is a prototype of the FTDI driver in action using luma.oled - a 2.42" SSD1309 connected via the FT232H to an internal USB port, mounted behind a piece of smoked acrylic in my workstation:

IMG_20190329_164157

thijstriemstra:
Sweet jezus! Want want want

rm-hull:
@thijstriemstra  - SPI over FDTI against SSD1309 SSD clocks up the following on perfloop:

```
#### iter =   2697: render time = 1.88 ms, frame rate = 532.59 FPS
```

Somewhat quicker than an RPi !

thijstriemstra:
haha nice work!

BeauSlim:
Very nice.  It occurred to me to try this with a Pi Zero in gpiozero "expander" mode, but it looked like a lot of work.  
Be careful depending on FTDI clones.  FTDI has a history of being really nasty with (Windows) drivers purposely bricking knock-off chips.

rm-hull:
Never heard of gpiozero expander mode before .. sounds interesting but like you say, seems like a lot of work.

This PR uses pyftdi to bridge over the FT232H, and that library only works on python versions >= 3.5... At some point we should stop supporting 2.7 soon anyway, especially since it is officially deprecated in 2020.

Anyway, I have SPI and GPIO working over FTDI and have tested with OLED devices and LED matrices. I am having some small troubles getting I2C working properly (naively I thought this would be the easiest!).

>  FTDI has a history of being really nasty with (Windows) drivers purposely bricking knock-off chips.

Ouch! Good to know... I can understand they want to protect their IP, but this seems quite evil.

thijstriemstra:
can you link the specific one you bought (from ebay)? I'll also get a few then.

rm-hull:
This link : https://rover.ebay.com/rover/0/0/0?mpre=https%3A%2F%2Fwww.ebay.co.uk%2Fulk%2Fitm%2F192189666695

rm-hull:
Ok, so I have a working I2C wrapper, but it is very slow - 0.16 FPS for an SSD1327 as measured by perfloop.py (I don't know why its so bad at the moment).

To get it working, there is a small hardware circuit needed: you need to tie the **AD1** and **AD2** pins together with a jumper wire.  One of the pins will be read by the F232H as an input, and the other pin will be used as an output.  When tied together both these pins form the **SDA** data line for I2C. The **AD0** pin alone will be the **SCL** clock line for I2C.

Also it is necessary to explicitly add **4.7KΩ** pull-up resistors from **SDA** & **SCL** lines up to **3V3**. This is necessary because the FT232H does not have pull-up resistors built in to these lines (whereas these are not necessary on the RPi as it does have pull-up resistors built-in).

image

rm-hull:
## Linux Installation Notes

1. Install native dependencies as follows:
```
sudo apt install libusb-1.0
```

2. Create a _udev_ configuration file to allow user-space processes to access
   the FTDI device. Typically, create a file, _/etc/udev/rules.d/11-ftdi.rules_:
```
# /etc/udev/rules.d/11-ftdi.rules
SUBSYSTEM=="usb", ATTR{idVendor}=="0403", ATTR{idProduct}=="6001", GROUP="plugdev", MODE="0666"
SUBSYSTEM=="usb", ATTR{idVendor}=="0403", ATTR{idProduct}=="6011", GROUP="plugdev", MODE="0666"
SUBSYSTEM=="usb", ATTR{idVendor}=="0403", ATTR{idProduct}=="6010", GROUP="plugdev", MODE="0666"
SUBSYSTEM=="usb", ATTR{idVendor}=="0403", ATTR{idProduct}=="6014", GROUP="plugdev", MODE="0666"
SUBSYSTEM=="usb", ATTR{idVendor}=="0403", ATTR{idProduct}=="6015", GROUP="plugdev", MODE="0666"
```

3. Add your user to the _plugdev_ group (log out and back in again to get the
   command to be effective):
```
sudo adduser $USER plugdev
```

4. Plug the device in and check `dmesg` output for the following:

```
[    3.192078] usb 1-4: new high-speed USB device number 2 using ehci-pci
[    3.360816] usb 1-4: New USB device found, idVendor=0403, idProduct=6014, bcdDevice= 9.00
[    3.360820] usb 1-4: New USB device strings: Mfr=1, Product=2, SerialNumber=0
[    3.360822] usb 1-4: Product: Single RS232-HS
[    3.360824] usb 1-4: Manufacturer: FTDI
[   45.472448] usbcore: registered new interface driver ftdi_sio
[   45.472469] usbserial: USB Serial support registered for FTDI USB Serial Device
[   45.472555] ftdi_sio 1-4:1.0: FTDI USB Serial Device converter detected
[   45.472600] usb 1-4: Detected FT232H
[   45.560372] usb 1-4: FTDI USB Serial Device converter now attached to ttyUSB0
```

## SPI Pin-outs

| OLED Pin | OLED Name | FT232 Pin | FT232 Function | Remarks |
|----------|----------|---------|----------|-------|
| 1 | GND | GND | GND | Ground |
| 2 | VCC | +3.3V | 3V3 | +3.3V Power |
| 3 | SCL | AD0 | SCLK | Serial Clock |
| 4 | SDA | AD1 | MOSI | Serial Data |
| 5 | RES | AD6 | GPIO 6 | Reset |
| 6 | DC | AD5 | GPIO 5 | Data/Command |
| 7 | CS | AD3 | CE0 | Chip Select |

## I2C Pin-outs

I2C devices need an additional wiring as documented [here](https://github.com/rm-hull/luma.core/pull/159#issuecomment-483800920).

| OLED Pin | OLED Name | FT232 Pin | FT232 Function | Remarks |
|----------|----------|---------|----------|-------|
| 1 | GND | GND | GND | Ground |
| 2 | VCC | +3.3V | 3V3 | +3.3V Power |
| 3 | SCL | AD0 | SCL | Serial Clock - connect to VCC rail via 4.7K pull-up resistor |
| 4 | SDA | AD1 + AD2 | MOSI | Serial Data - connect to VCC rail via 4.7K pull-up resistor |

rm-hull:
@thijstriemstra if you could review please. I'm not going to explicitly add RST documentation for these classes here - but _will_ try and make a concerted effort to pull all the documentation for _all_ modules into https://github.com/rm-hull/luma.docs instead. 

We also need to look into why the readthedocs doesn't appear to be auto-generating documentation from master anymore.

CHANGES.rst
```diff
@@ -5,6 +5,7 @@ ChangeLog
 | Version    | Description                                                         | Date       |
 +============+=====================================================================+============+
 | *upcoming* | * Improve I2C render speed by not chunking into 32-byte blocks      | TBC        |
+|            | * Add support for bridging FTDI to I2C/SPI                          |            |
```
thijstriemstra: I wouldn't mind calling this release 2.0

luma/core/interface/ftdi.py
```diff
@@ -0,0 +1,63 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2019 Richard Hull
+# See LICENSE.rst for details.
+
+
```
thijstriemstra: docstring inc version added info is missing.

luma/core/interface/serial.py
```diff
@@ -295,3 +295,84 @@ def __setattr__(self, attr, val):  # pragma: no cover
 
     def __noop(self, *args, **kwargs):
         pass
+
+
+def ftdi_spi(device='ftdi://::/1', bus_speed_hz=12000000, CS=3, DC=5, RESET=6):
```
thijstriemstra: i'd prefer `cs_pin`, `dc_pin` etc, or at least lowercase them.

setup.py
```diff
@@ -37,7 +37,8 @@ def find_version(*file_paths):
 
 install_deps = [
     'pillow>=4.0.0',
-    'smbus2'
+    'smbus2',
+    'pyftdi;python_version>="3.5"'
```
thijstriemstra: this is an optional dependency.. looking at pyftdi, it depends on

```
INSTALL_REQUIRES = [
    'pyusb >= 1.0.0',
    'pyserial >= 3.0',
]
```

so installing luma.core would also pull in these dependencies. I'd rather not do that.. thoughts?

luma/core/interface/serial.py
thijstriemstra: and can these methods not also be moved to `luma/core/interface/ftdi.py`? and import them here?

setup.py
rm-hull: Surely pyftdi is optional in the same way smbus or spidev are both optional - if you only want to run an I2C display, it still installs spidev for example.

luma/core/interface/ftdi.py
rm-hull: These aren't meant to be for public consumption, so I might just inline them into the serial.py file

CHANGES.rst
rm-hull: Major version number change implies a breaking change though - which this is not.

luma/core/interface/serial.py
rm-hull: Have used `gpio_CS`, `gpio_DC`, `gpio_RST` to align with same parameter names used by the SPI serial interface

CHANGES.rst
thijstriemstra: a breaking change or a big new feature (which this is), but yea, fine with me.

setup.py
thijstriemstra: > if you only want to run an I2C display, it still installs spidev for example.

if you only want to run an i2c display, it will install pyusb and pyserial that you do not need. every luma library (depends on core) will now install pyserial and pyusb that you do not need..

rm-hull: So what is the alternative? Not pyftdi a dependency and at runtime cause luma.core to crash with an import error? That makes for a crappy user experience and an extra burden on fielding github issues, for the only benefit of a few less files that were installed as a transitive dependency.

One could argue that there are many libraries installed (both at the OS level, and within the python installation directories) that you do not need, but they are installed anyway. 

We aren’t talking about devices that are particularly memory- or storage-constrained, and if you are concerned about polluting your main python dist-packages or site-packages with unnecessary cruft, then surely there’s virtualenv/pipenv/etc to manage all that.

I think I’m just unsure what the hesitation is about pulling in those dependencies verses the trade-offs of not doing so, ... and then the complexity of having to manage installing them somewhere else.

luma/core/interface/ftdi.py
thijstriemstra: the fact they're not meant for public consumption does not mean it's still useful to know when they were added IMO.

setup.py
thijstriemstra: > Not making pyftdi a dependency and at runtime cause luma.core to crash with an import error?

I wouldn't call it a crash tbh. It's like the luma.examples that look for psutil and warn if it's not installed. I prefer to keep things simple and lightweight and installing a dependency that is only used for a small part of luma.core feels like overkill. Adding these dependencies now (without a small 'fight') means we will keep bloating the library. This issue reminds me of the pygame dependency back in the day. We eventually separated it into a different library (luma.emulator), making it trivial to install luma.core/luma.oled because installing pygame (without an egg) is not easy and we didn't want to put this uncessary burden on the user if they do not even use it. I have a similar feeling with this. These dependencies could result in install hell. And what are we going to tell the user then, something like:

- git clone the repo
- manually remove the pyftdi dep from setup.py
- and `pip install -e .`?

maybe I am exagerating but less is more IMO.

What happens if you do a pyftdi installation in a clean RPI system? If it doesn't cause any issues, I am +1 for including this dep. If it causes issues we have to take a look at those because adding extra complexity to luma.core (which will affect all other luma libs) sounds like a really bad idea to me.

rm-hull: > installing pygame (without an egg) is not easy

Good point .. had forgotten about that on RPi

> What happens if you do a pyftdi installation in a clean RPI system?

Will try tomorrow

thijstriemstra: >  if you are concerned about polluting your main python dist-packages or site-packages with unnecessary cruft

it's more that you first need to `apt-get` a bunch of stuff before you can properly pip install some packages.

> Will try tomorrow

Thanks.

rm-hull:
`libusb-1.0` was already installed via apt on my RPi B2 with Debian Jessie.

Also, it looks like pyserial was already installed under python 3:

```
$ pip3 install pyftdi
Downloading/unpacking pyftdi
  Downloading pyftdi-0.29.4.tar.gz (66kB): 66kB downloaded
  Running setup.py (path:/tmp/pip-build-05nj576q/pyftdi/setup.py) egg_info for package pyftdi
Downloading/unpacking pyusb>=1.0.0 (from pyftdi)
  Downloading pyusb-1.0.2.tar.gz (54kB): 54kB downloaded
  Running setup.py (path:/tmp/pip-build-05nj576q/pyusb/setup.py) egg_info for package pyusb
    warning: no files found matching 'ReleaseNotes.rst'
Downloading/unpacking pyserial>=3.0 (from pyftdi)
  Downloading pyserial-3.4-py2.py3-none-any.whl (193kB): 193kB downloaded
Installing collected packages: pyftdi, pyusb, pyserial
  Running setup.py install for pyftdi
    changing mode of build/scripts-3.4/i2cscan.py from 644 to 755
    changing mode of /usr/local/bin/i2cscan.py to 755
  Could not find .egg-info directory in install record for pyftdi
  Running setup.py install for pyusb
    warning: no files found matching 'ReleaseNotes.rst'
  Could not find .egg-info directory in install record for pyusb>=1.0.0 (from pyftdi)
  Found existing installation: pyserial 2.6
    Not uninstalling pyserial at /usr/lib/python3/dist-packages, owned by OS
Successfully installed pyftdi pyusb pyserial
Cleaning up...
```

thijstriemstra:
> libusb-1.0 was already installed via apt

you mean, you installed it or it comes with jessie? also, jessie is ancient and knowing if it's part of latest raspbian would be useful.

rm-hull:
On Raspian stretch:

```
$ apt list libusb-1.0-0
Listing... Done
libusb-1.0-0/stable,now 2:1.0.21-1 armhf [installed,automatic]
```

rm-hull:
v1.9.0 released to pypi with this change incorporated

TCB13:
What's the likelihood / instructions to this it to work with the FT2232H? Adafruit's Blinka already supports FT2232H so is there anything else that is needed to be changed in Luma? Thank you.

rm-hull:
Would probably better directing this question (initially at least) to the author's of pyftdi and linking back their answer. There are no immediate plans to support that unless it is trivially simple.

TCB13:
> Would probably better directing this question (initially at least) to the author's of pyftdi and linking back their answer. There are no immediate plans to support that unless it is trivially simple.

According to this https://eblot.github.io/pyftdi/#overview:

> Supported FTDI devices include:
> (...)
> FT2232H (dual port, clock up to 30 MHz)
> FT4232H (quad port, clock up to 30 MHz)

They are supported.

rm-hull:
In which case, when you initialise the SPI using https://luma-core.readthedocs.io/en/latest/interface.html#luma.core.interface.serial.ftdi_spi (or equivalent for I2C), just specify the device as `ftdi://ftdi:2232h/1` rather than leaving it blank

TCB13:
So that means I should use the `luma.core.interface.serial.ftdi_spi` class instead of the generic SPI `luma.core.interface.serial.spi` like the example here https://github.com/rm-hull/luma.oled/issues/185 ? Is that right?

``luma.core.interface.serial.spi` also depends on `Adafruit_GPIO` and they don't have a definition for FT2232H.

Thank you.

rm-hull:
This is how I initialized in an another project: https://github.com/rm-hull/TR4-monitor/blob/master/tr4_monitor/main.py#L172-L174

You will want to do something similar, but override the default FTDI device:

```python
from luma.oled.device import ssd1309
from luma.core.interface.serial import ftdi_spi
return ssd1309(ftdi_spi(device='ftdi://ftdi:2232h/1'), **vars(args))
```

PR #161, opened 2019-05-22
Allow viewport pseudo-display to dither when rendering to device with lower-fidelity color mode 

rm-hull:


luma/core/virtual.py
```diff
@@ -38,15 +38,33 @@ def range_overlap(a_min, a_max, b_min, b_max):
 
 
 class viewport(mixin.capabilities):
-
-    def __init__(self, device, width, height):
-        self.capabilities(width, height, rotate=0, mode=device.mode)
+    """
+    The viewport offers a positionable window into a larger resolution pseudo-display,
+    that also supports the concept of hotspots (which act like live displays).
+    :param device: The device to project the enlarged pseudo-display viewport onto.
```
thijstriemstra: add a newline between description and params

rm-hull: Done

rm-hull:
:+1:

PR #162, opened 2019-05-24
Chunk `smbus.i2c_rdwr()` calls into 4Kb blocks to avoid ioctl error 22

rm-hull:
Fixes https://github.com/rm-hull/luma.oled/issues/253


luma/core/interface/serial.py
```diff
@@ -103,14 +103,25 @@ def data(self, data):
         :type data: list, bytearray
         """
         if self._managed:
-            self._bus.i2c_rdwr(self._i2c_msg_write(self._addr, [self._data_mode] + data))
+            block_size = 4096
```
thijstriemstra: can you add a comment for `block_size` describing why this magic nr is needed?

rm-hull: Added comment - I started out with a block size of 1K as the ioctl docs alluded to, then kept doubling it until it no longer worked. 4K is the last block size that doesn't fail on RPi. It seems a good compromise between transferring large chunks and fewer calls.

PR #167, opened 2019-07-09
use latest 4.x release of pytest

thijstriemstra:


thijstriemstra:
weird it also fails on pytest 5.x because locally it seems to work fine (besides 1 test failing as usual);

```
reference = <PIL.PngImagePlugin.PngImageFile image mode=RGB size=128x64 at 0x748B6290>
target = <PIL.Image.Image image mode=RGB size=128x64 at 0x748B63F0>
img_path = '/home/pi/projects/pi-test/luma.core/tests/reference/images/accented_charset.png'

    def assert_identical_image(reference, target, img_path):
        bbox = ImageChops.difference(reference, target).getbbox()
        assert bbox is None, '{0} is not identical to generated image'.format(
>           os.path.basename(img_path))
E       AssertionError: accented_charset.png is not identical to generated image

tests/helpers.py:56: AssertionError
======================================== 1 failed, 149 passed in 11.32 seconds =========================================
```

rm-hull:
I think pillow 6.1.0 has been released recently (within the last week) - this is what has caused the image test to fail its assertions.

thijstriemstra:
pytest 4.6 will be last to support Python 2.7. Since we don't plan to drop Python 2.7 support this soon, it's probably best to just stick to pytest 4.5 as it is now. Closing PR.

See https://docs.pytest.org/en/latest/py27-py34-deprecation.html

PR #172, opened 2019-12-20
Add support for using any GPIO pin as a CS pin

### LINKED ISSUE #170: Support arbitrary GPIO pins for SPI chip select
The goal would be to support more than 2 devices.

I don't mind implementing this and submitting a PR, I'm just wondering if this is something you'd accept or if you think the library should be limited to what spidev supports.

The idea would be to initialize spidev with [`no_cs`](https://github.com/doceme/py-spidev#settings) and control the chip select with GPIO from luma itself (either always or only when an override option is set to `true`, thoughts on what's best?)

ric2b:
fixes #170

There seems to be a bug with spidev where it ignores the `no_cs` option: https://github.com/doceme/py-spidev/issues/46

Besides that, I've tested it locally with 3 screens on a single `(port, device)` and it works as expected.

### DIFF
--- a/luma/core/interface/serial.py
+++ b/luma/core/interface/serial.py
@@ -280,9 +280,9 @@ class spi(bitbang):
             self._spi.open(port, device)
             self._spi.cshigh = cs_high
             if gpio_CS:
-                self._spi.no_cs = True # disable spidev's handling of the chip select pin
+                self._spi.no_cs = True  # disable spidev's handling of the chip select pin
                 self._cs_high = cs_high
-                
+
         except (IOError, OSError) as e:
             if e.errno == errno.ENOENT:
                 raise luma.core.error.DeviceNotFoundError('SPI device not found')
@@ -295,9 +295,9 @@ class spi(bitbang):
         gpio = self._gpio
         if self._CE:
             gpio.output(self._CE, gpio.HIGH if self._cs_high else gpio.LOW)
-        
+
         self._spi.writebytes(data)
-        
+
         if self._CE:
             gpio.output(self._CE, gpio.LOW if self._cs_high else gpio.HIGH)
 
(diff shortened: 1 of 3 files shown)


thijstriemstra:
@ric2b can you also add a unit test for this?

luma/core/interface/serial.py
```diff
@@ -270,15 +270,19 @@ class spi(bitbang):
     """
     def __init__(self, spi=None, gpio=None, port=0, device=0,
                  bus_speed_hz=8000000, cs_high=False, transfer_size=4096,
-                 gpio_DC=24, gpio_RST=25):
+                 gpio_CS=None, gpio_DC=24, gpio_RST=25):
```
thijstriemstra: why not add this as last argument? i'd prefer that

ric2b: I was just going with the order in the bitbang class, but either way is fine for me, changed :)

ric2b:
> @ric2b can you also add a unit test for this?

Done, but let me know if you think I should cover more cases

rm-hull:
Also: I couldn't see where `self._CE` was being set ?

ric2b:
> Does that seem reasonable?

Yup, I'll make the changes :) But I'd prefer to call it `gpio_cs_spi` or something similar instead of `nocs_spi`, I think the latter might hides the fact that this controls the cs pin, just in a different way.

ric2b:
> Also: I couldn't see where `self._CE` was being set ?

It's done [here](https://github.com/rm-hull/luma.core/pull/172/files#diff-c6ec9436479802f3b77a187887e21314R278), since the `bitbang` class [gets it from kwargs](https://github.com/rm-hull/luma.core/blob/master/luma/core/interface/serial.py#L173).

ric2b:
The last commits about python 2, let me know if you have good alternatives, I don't love doing a pop on kwargs :/

(and sorry for the delay, I've been working on other things)

PR #173, opened 2020-02-28
Elide python 2.7 from setup & project infra

rm-hull:
Not sure why we cant use latest pytest (as per 1st commit) - not really had chance to look into this failure: https://travis-ci.org/rm-hull/luma.core/builds/656037029

### DIFF
--- a/tests/test_ftdi_spi.py
+++ b/tests/test_ftdi_spi.py
@@ -11,10 +11,9 @@ import pytest
 from unittest.mock import Mock, call, patch
 import sys
 from luma.core.interface.serial import ftdi_spi
-from helpers import pyftdi_missing, fib
+from helpers import fib
 
 
-@pytest.mark.skipif(sys.version_info < (3, 5), reason=pyftdi_missing)
 @patch('pyftdi.spi.SpiController')
 def test_init(mock_controller):
     gpio = Mock()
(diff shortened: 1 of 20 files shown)


rm-hull:
I can't find a specific reference to py27 in the readme...
Do you mean this badge: image

If so, that would get updated as soon as a new release goes out.

thijstriemstra:
> I can't find a specific reference to py27 in the readme...

You're right, was confused with other sub-projects. There are no Python references at all though, perhaps you can add `Python 3`? e.g.

```
luma.core is a component library providing a Pillow-compatible drawing canvas for Python 3, 
```

thijstriemstra:
Looks good. I wonder if a note should be added that version X.X is the last to support Python 2.7, I guess it could go in the changelog.

tests/test_ftdi_spi.py
```diff
@@ -8,9 +8,10 @@
...
 
 import pytest
+from unittest.mock import Mock, call, patch
 import sys
 from luma.core.interface.serial import ftdi_spi
-from helpers import Mock, call, patch, pyftdi_missing, fib
+from helpers import pyftdi_missing, fib
 
 
 @pytest.mark.skipif(sys.version_info < (3, 5), reason=pyftdi_missing)
```
thijstriemstra: this check for < 3.5 can go I guess.

thijstriemstra:
And there's also https://github.com/rm-hull/luma.core/blob/master/.readthedocs.yml

but I don't remember why it was pinned to python 2.7..

rm-hull:
Thought this was going to be a quickie PR ... turns out that’s not the case!

PR #179, opened 2020-07-21
snapshot: always redraw first should_redraw

baslr:
If you have a snapshot with interval 60 seconds and the software runs immediately after system start the first redraw does not happen.
```
#           20 - 0 > 60 is false
return time.monotonic() - self.last_updated > self.interval
```

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -30,3 +30,4 @@ Contributors
 * Daniel Smullen (@drspangle)
 * Hans Liss (@hansliss)
 * Phil Howard (@gadgetoid)
+* Manuel Baesler (@baslr)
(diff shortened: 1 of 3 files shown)


rm-hull:
There ought to be an assert at the top of the constructor too, e.g.:

```python
assert interval > 0
```

baslr:
> That’s great ... could you add your details to the contributors list too please (https://github.com/rm-hull/luma.core/blob/master/CONTRIBUTING.rst)

@rm-hull done.

rm-hull:
I’ll build a new release shortly with these changes in. 
Thank you for the PR 🙂

rm-hull:
@baslr release 1.14.1 was released with your changes 🚀

PR #189, opened 2020-09-20
doc: improve api docs

thijstriemstra:


luma/core/interface/serial.py
```diff
@@ -525,7 +525,7 @@ class pcf8574(i2c):
     :param BACKLIGHT: value aligned with backlight pin (default: 3)
```
thijstriemstra, line 525: @dhrone what do you mean with value aligned? simply the backlight pin?

dhrone: Value aligned as in which bit within a byte is aligned to the pin associated with the backlight.

The PCF8574 has 8 data ports (p0-p7) which are set based upon the byte that you are sending.  The bits that are set within the byte determine which data ports are set high or low.

```
dataline:         p7  p6  p5  p4 p3 p2 p1 p0
value:           128, 64, 32, 16, 8, 4, 2, 1
```

So, for instance, if you were to send a 0x80 (128), p7 would be set high with the remaining pins set low.  Sending a 0x41 (65) would set data ports p6 and p0 high and the remainder low. 

It is completely arbitrary which data lines are connected to which signal lines on the display.  With the display that I am using, the backlight just happens to be connected to p3.  It seemed to make sense for the driver to describe the pins based upon the bitwise location as opposed to their binary value. 

So, if the PCF8574 pin that is connected to the backlight is p3, we need to 'or' an 8 with the value that we are sending to ensure that p3 is set high.

That is why the pin value is computed with the following formula:
`_pin = 1 << pin`
where `pin` is the number of the pin (starting from zero) that the backlight is connected to.

This value is then or'd with any subsequent bytes sent to the PCF8574.

dhrone (approved):

luma/core/interface/serial.py
dhrone: Put more simply, param BACKLIGHT should equal the pin number of the pcf8574 (counting from zero) that the backlight is controlled from.

thijstriemstra: thanks for feedback, made change.

dhrone (approved):

PR #190, opened 2020-09-22
Allow reset to be held active for a configurable duration

### LINKED ISSUE #187: LCD ST7735 128X128  - WHITE SCREEN
Hello guys how are you all doing? Its my first time ever using GitHub so I'm sorry if I'm not using the right place to find an answer to my problem. 

#### Type of Raspberry Pi and LCD specifications
Raspberry Pi 3B+
ST7735 LCD SPI 1.44 128x128

Place where I bought the LCD:
[https://www.aliexpress.com/item/GREATZT-1PCS-5V-3-3V-1-44-inch-TFT-LCD-Display-Module-128-128-Color-Sreen/32948378014.html?spm=a2g0s.9042311.0.0.1f614c4dALAgbo](url)

#### Linux Kernel version
Linux retropie 4.14.79-v7+ #1159 SMP Sun Nov 4 17:50:20 GMT 2018 armv7l GNU/Linux
Library: luma.lcd 1.1.0 (luma.core 1.8.3)
Python-dev: 2.7.13-2
Python-pip: 9.0.1-2+rpt2

#### Expected behaviour
I am willing to use this LCD to show at full time all the status of my Raspberry. Such as the cpu temp, storage, IP address and etc. My plan is to make this script start up as soon as my rasp starts too. So I created a sh file to be triggered by the rc.local all the time it boots up. 

My gpio set up is just like the library document:
led 18
sck 11(sclk)
sda 10(mosi)
ao 23
reset 24
cs 8(ceo)
gnd
vcc 3.3/5

This is how it looks whenever it works:
image

This is my status script: https://paste.ofcode.org/LuY4CDcwSmMcXjGzUBjXx

This is the sh script witch the rc.local triggers when the rasp boots up:
image

#### Actual behaviour
My actual problem is that sometimes the LCD will turn on but completely white. Even if you check it closer you won't be able to see anything or even blinking. 

image

So, to make it work I go to the command line and and input:

image

Even running this command in the manual way the screen will shows itself all white, so I have to interrupt and run again until it shows the status screen as I showed before.

I tried to play with those commands from the library such as the device.clean, decive.show however I will randomly get this white screen. 

Have you ever seen this type of issue? 

Note: I tried to change from the 3.3 to 5.0 but the problem is still the same.
Note_2: I also tried to wait a while to see if it would change but nothing happens the white screen will stay up until you interrupt.

In case I was not clear please let me know so I can even make a video. Thanks for this amazing library and also for the attention.

rm-hull:
Fixes #187 

Provides an optional parameter (and cmd-line switch) to prolong the reset hold duration. It seems like some devices need the reset line to be kept active for a minimum period.

Usage would be like:
```python
serial = spi(device=0, port=0, reset_hold_time=0.1)  # 100ms
device = st7735(serial)
```

If the parameter is not provided, existing behaviour is not changed.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,7 @@ ChangeLog
 | Version    | Description                                                         | Date       |
 +============+=====================================================================+============+
 | *TBC*      | * Added --inverse option for ST7735 to cmdline opt                  | TBC        |
+|            | * Allow SPI interface to define reset duration                      |            |
 +------------+---------------------------------------------------------------------+------------+
 | **1.16.2** | * Added new parallel interface module                               | 2020/09/20 |
 |            | * Renamed parallel class to bitbang_6800; moved to parallel module  |            |
--- a/luma/core/cmdline.py
+++ b/luma/core/cmdline.py
@@ -156,6 +156,7 @@ class make_serial(object):
                    cs_high=self.opts.spi_cs_high,
                    transfer_size=self.opts.spi_transfer_size,
                    reset_hold_time=self.opts.gpio_reset_hold_time,
+                   reset_release_time=self.opts.gpio_reset_release_time,
                    gpio_DC=self.opts.gpio_data_command,
                    gpio_RST=self.opts.gpio_reset,
                    gpio=self.gpio or GPIO)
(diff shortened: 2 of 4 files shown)


dhrone (dismissed):
According to the ST7735 documentation, a reset is triggered if RESX is held low longer than 5us and can take up to 120ms to complete depending on the power state of the display (see section 9.16 in [ST7735 data sheet](https://www.displayfuture.com/Display/datasheet/controller/ST7735.pdf))  If this were a display specific setting (as opposed to the interface), I'd recommend making the default something like 150ms.  Since it is not, this seems like a reasonable approach.  

FYI, it is possible that the problem is not how long you held the pin low as I doubt that you can toggle the pin from high-low-high in under 5us using RPi.GPIO. I suspect instead it was beginning to transmit data before the reset was completed.  This issue of certain commands (like reset) taking longer than other commands is a problem I was attempting to solve by introducing an optional exec_time in parallel_device so that depending on the command being issued, an appropriate delay to wait for the command to finish could be specified.  Just something to think about...

One other item, if you are going to release another version of `luma.core`, I'd like to introduce a change to the exception handling of `bitmap_font` to eliminate the dependency on `Pillow` >=7.  I'll put a PR up for that shortly.  Happy to pull it back if you want to take a different direction.

rm-hull:
> One other item, if you are going to release another version of luma.core, I'd like to introduce a change to the exception handling of bitmap_font to eliminate the dependency on Pillow >=7. I'll put a PR up for that shortly. Happy to pull it back if you want to take a different direction.

I merged it

rm-hull:
ok, so maybe there should be an additional sleep after setting reset low

thijstriemstra (approved):

PR #192, opened 2020-09-25
Add support for additional --interface options: "pcf8574" and "bitbang_6800"

rm-hull:


luma/core/cmdline.py
```diff
@@ -121,7 +121,7 @@ def load_config(path):
     return args
 
 
-class make_serial(object):
+class make_interface(object):
```
thijstriemstra, line 124: this is a backward incompat change. Can you add a `make_serial = make_interface` with a comment it's deprecated?

rm-hull: done

thijstriemstra (approved):
Looks good to me but a new release of luma.lcd (and luma.oled?) needs to be released first for this work?

thijstriemstra:
ps. travis build is failing

rm-hull:
Don’t think a new release of oled or lcd packages are necessary. This is just about exposing the interfaces (in luma.core) as command line args. At most, it will require a bump in luma.examples only.

rm-hull:
Btw, I think there is another PR in the making, to tidy up `serial_interface` into just `interface` in the device classes. I will retain the old parameter name but mark it as deprecated, but this could have quite a ripple effect through all the packages ...

thijstriemstra (approved):

PR #194, opened 2020-09-27
Linux /dev/fbX framebuffer display device

rm-hull:
### What is this for?
To allow programs that use luma.core classes to write directly to the Linux framebuffer

### Why?
1. A pure python implementation of the display logic is the limiting factor in terms of performance for devices with larger screen resolutions (e.g. ili9341 with 320x240 pixels).
2. Every new device class needs someone to write custom code and test the display driver for it.

### How is this PR the solution?
If there are pre-existing kernel framebuffer drivers for a wide range of devices (and there are are: https://git.kernel.org/pub/scm/linux/kernel/git/gregkh/staging.git/tree/drivers/staging/fbtft?h=staging-testing) which are already highly performant (on account of being written in C), we can leverage that by just dumping the image contents to the relevant `/dev/fbN` framebuffer device, and rely on the kernel to flush the image out to the display.

This allows us to continue using the regular luma.core library methods across a range of devices, and be able to support devices with larger screen resolutions. The only caveat being that the current user must be added to the relevant user group:

```console
$ sudo adduser ${USER} video
```

### TODO Actions
- [x] Add class/method docstrings for  the `fb` device (& add `.. versionadded::` tag)
- [ ] Include documentation/usage example
- [x] Update changelog 
- [x] Is luma.core the correct package for this new device (luma.emulator instead?)
- [x] Is the approach to the problem appropriate?
- [x] Can anything be simplified?
- [x] Any obvious security flaws or potential holes?
- [x] Is there anything included that should not be?

### Interim documentation
The following was tried with an SPI-base ~~SSD1351~~ ST7735R display.

**IMPORTANT NOTE:** Ensure the kernel is updated to the latest with `sudo rpi-update`. This was tested with kernel version as below:
```
$ uname -a
Linux rpi-b4-8M 5.4.69-v7l+ #1344 SMP Fri Oct 2 16:01:56 BST 2020 armv7l GNU/Linux
```

First connect up the device according to: https://luma-oled.readthedocs.io/en/latest/hardware.html#spi, then edit `/boot/config.txt` to include the following line (I added it under the `[all]` section:

```
dtoverlay=adafruit18,dc_pin=23,reset_pin=24
```

Obviously, set the pin numverss to the correct GPIO BCM pins. The full list of overlay options are documented in file `/boot/overlays/README`, and it appears there are drivers for OLEDs SH1106, SSD1306 (both SPI and I2C), SSD1351, as well as other LCDs.

Then reboot the RPi and check that it has loaded the framebuffer kernel driver correctly:

```console
$ dmesg | grep fb
[    1.005683] pci_bus 0000:00: root bus resource [mem 0x600000000-0x603ffffff] (bus address [0xf8000000-0xfbffffff])
[    4.901643] fbtft: module is from the staging directory, the quality is unknown, you have been warned.
[    4.906686] fb_st7735r: module is from the staging directory, the quality is unknown, you have been warned.
[    4.907366] fbtft_of_value: width = 128
[    4.907382] fbtft_of_value: height = 160
[    4.907399] fbtft_of_value: buswidth = 8
[    4.907418] fbtft_of_value: debug = 0
[    4.907433] fbtft_of_value: rotate = 90
[    4.907449] fbtft_of_value: fps = 50
[    5.859750] graphics fb0: fb_st7735r frame buffer, 160x128, 40 KiB video memory, 4 KiB buffer memory, fps=50, spi0.0 at 40 MHz


$ ls -l /dev/fb0 
crw-rw---- 1 root video 29, 0 Oct  5 17:28 /dev/fb0

$ cat /sys/class/graphics/fb0/virtual_size 
160,128

$ cat /sys/class/graphics/fb0/bits_per_pixel 
16
```
Note, if you are also connected to an HDMI monitor, the framebuffer for the small display may be at `/dev/fb1` or `/dev/fb2`. On rebooting the console output should go to the display:

image


luma/core/device.py
```diff
@@ -10,6 +11,8 @@
 import luma.core.const
 from luma.core.interface.serial import i2c, noop
 
+__all__ = ["fb"]
```
thijstriemstra, line 14: doesn't this limit the imports to just `fb`; and thus does this not break importing other classes from this module?

luma/core/device.py
```diff
@@ -168,6 +173,62 @@ def display(self, image):
         :param image: Image to display.
         :type image: PIL.Image.Image
         """
-        assert(image.size == self.size)
+        assert image.size == self.size
 
         self.image = self.preprocess(image).copy()
+
+
+class fb(device):
```
thijstriemstra, line 181: missing docstring and version marker.

luma/core/device.py
```diff
@@ -168,6 +173,62 @@ def display(self, image):
...
+        super(fb, self).__init__(serial_interface=noop())
+        self.id = self.__get_display_id(display)
+        (width, height) = self.__config("virtual_size")
+        self.bits_per_pixel = next(self.__config("bits_per_pixel"))
+        assert self.bits_per_pixel in [24, 32]
+        self.capabilities(width, height, rotate, mode="RGB")
+
+    def __get_display_id(self, display):
+        """
+        Allow for :1 or /dev/fb1, etc
```
thijstriemstra, line 192: Can you elaborate and expand this docstring.

luma/core/device.py
```diff
@@ -168,6 +173,62 @@ def display(self, image):
...
+
+        if display.startswith("/dev/fb"):
+            return int(display[7:])
+
+        raise luma.core.error.DeviceNotFoundError(
+            "Invalid/unsupported framebuffer: {}".format(display)
+        )
+
+    def __config(self, section):
+        path = "/sys/class/graphics/fb{0}/{1}".format(self.id, section)
```
thijstriemstra, line 208: can you move this string to a class var/constant?

tests/helpers.py
```diff
@@ -11,27 +11,27 @@
...
+    platform.system()
+)
 
 
 def get_reference_file(fname):
-    return os.path.abspath(os.path.join(
-        os.path.dirname(__file__),
-        'reference',
-        fname))
+    return os.path.abspath(os.path.join(os.path.dirname(__file__), "reference", fname))
```
thijstriemstra, line 26: can you use `pathlib` here instead? Since we're py3 only now.

tests/helpers.py
```diff
@@ -11,27 +11,27 @@
...
 
 
 def get_reference_image(fname):
-    return get_reference_file(os.path.join('images', fname))
```
thijstriemstra, line 30: replace os.path usage with pathlib

tests/test_fb_device.py
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# Copyright (c) 2017-18 Richard Hull and contributors
```
thijstriemstra, line 3: only 2020

tox.ini
```diff
@@ -2,16 +2,20 @@
...
 
 [testenv]
 commands =
     python setup.py install
-    coverage run -m py.test -v -r wsx -s
+    coverage run -m py.test -v -r wsx -s {posargs}
     coverage report
 deps = .[test]
 
+[testenv:watch]
```
thijstriemstra, line 15: what does watch do? can you add a comment here with short description?

luma/core/device.py
```diff
@@ -168,6 +173,62 @@ def display(self, image):
...
         self.image = self.preprocess(image).copy()
+
+
+class fb(device):
+    def __init__(self, display=None, rotate=0, **kwargs):
+        super(fb, self).__init__(serial_interface=noop())
+        self.id = self.__get_display_id(display)
+        (width, height) = self.__config("virtual_size")
+        self.bits_per_pixel = next(self.__config("bits_per_pixel"))
+        assert self.bits_per_pixel in [24, 32]
```
thijstriemstra, line 187: what are these magic nrs here? Can you assign a meaningful var name to this 24, 32?

luma/core/device.py
thijstriemstra: so perhaps move this new `fb` class to a separate (new) module?

rm-hull: I had to recheck, but according to https://docs.python.org/3/tutorial/modules.html#importing-from-a-package:

> ... if a package’s `__init__.py` code defines a list named `__all__`, it is taken to be the list of module names that should be imported when `from package import *` is encountered ...

Hence it should not break imports because we never do wildcard imports in any sub-modules, and end-user code should never have to import either the device, parallel_device or dummy classes.

luma/core/device.py
rm-hull: I would prefer to keep it localized to the method. There are no other usages in the class of the string, and it is specific to the config, so keeping it here seems to be the right thing to ensure it is encapsulated.

tox.ini
rm-hull: allows you to run `tox -e watch` which in turn runs pytestwatch which will automatically run tests when files change. 

Obvs. this is useful for development only.

luma/core/device.py
rm-hull: 24 relates to a bit depth of 24 - basically enough to store RGB colours in 3 bytes per pixel
32 is RGBA - the extra byte is the alpha channel, which needs 4 bytes per pixel

luma/core/device.py
thijstriemstra: thanks for looking into it, sounds good.

tox.ini
thijstriemstra: "use pytestwatch to run tests when files change" sounds like a good comment to me

luma/core/device.py
thijstriemstra: it's used in the tests multiple times, that's why I suggested it. Up to you.

rm-hull:
Just an FYI/update, I saw this in the fbtft wiki: https://github.com/notro/fbtft/wiki#fbtft_device-and-flexfb-are-gone-in-54
So this means there is quite some disruptive change which means a whole bunch of kernel framebuffer drivers will just stop working on the latest kernel. 

* I cannot get the `fb_ssd1351` kernel driver displaying anything on the screen despite it working perfectly from luma.oled.
* I tried a st7735r display, but cannot fathom the right incantation since `fbtft_device` is no more
* I can get the `fb_ssd1306` kernel driver displaying something (the boot console shows various services starting up), but it is painfully slow, but i assume this is because its over i2c .. the silver lining is that the `luma.core.device.fb` code appears to _nearly_ work (the `_toMono()` method need to be significantly refined): it is displaying something that looks recognisable as the shapes from the examples demo
* I have an ILI9325 somewhere, so i will see if i can press that into service which I think is still supported with the dtoverlay mechanism

luma/core/cmdline.py
```diff
@@ -131,6 +131,10 @@ def __init__(self, opts, gpio=None):
         self.opts = opts
         self.gpio = gpio
 
+    def noop(self):
```
thijstriemstra, line 134: i wouldn't call it noop (there are now 2 `noop`'s in pythons namespace, or overwriting eachother), just call it something like `get_noop`

rm-hull: i think it has to be called noop (look at all the other methods in the class: https://github.com/rm-hull/luma.core/blob/feature/framebuffer-device/luma/core/cmdline.py#L125-L214) as it gets instantiated dynamically (see https://github.com/rm-hull/luma.core/blob/feature/framebuffer-device/luma/core/cmdline.py#L241 for example, from the interface given).

Arguably, this should be in a separate PR, it was missing functionality that shouldn't really be in this PR ... i might move it out and do it separately actually

thijstriemstra: sounds good.

luma/core/device.py
thijstriemstra: to me, sounds like a good comment to add

luma/core/cmdline.py
```diff
@@ -300,14 +306,18 @@ def create_parser(description):
     ftdi_group = parser.add_argument_group('FTDI')
     ftdi_group.add_argument('--ftdi-device', type=str, default='ftdi://::/1', help='FTDI device')
 
+    linux_framebuffer_group = parser.add_argument_group('LINUX_FRAMEBUFFER')
```
thijstriemstra, line 309: 'Linux framebuffer' sounds/looks better to me.

setup.py
```diff
@@ -31,7 +31,8 @@ def find_version(*file_paths):
 pytest_runner = ['pytest-runner'] if needs_pytest else []
 test_deps = [
     "pytest==4.5",
```
thijstriemstra, line 33: still no luck with newest pytest/pytest-cov?

thijstriemstra: i tried again with various combos of pytest/pytest-cov; can't figure it out

rm-hull: Yeah, i had a try again at this, ... and gave up :(

luma/core/device.py
```diff
@@ -168,6 +173,89 @@ def display(self, image):
         :param image: Image to display.
         :type image: PIL.Image.Image
         """
-        assert(image.size == self.size)
+        assert image.size == self.size
 
         self.image = self.preprocess(image).copy()
+
+
+def _toRGB565(image):
```
thijstriemstra, line 181: these helpers belong in `luma.core.util`

tests/test_fb_device.py
```diff
@@ -0,0 +1,124 @@
...
+    ) as fake_open:
+        device = fb("/dev/fb1")
+        assert device.bits_per_pixel == 24
+        fake_open.assert_has_calls(
+            [call("/sys/class/graphics/fb1/bits_per_pixel", "r")]
+        )
+
+
+def test_display_1bpp():
+    with open("tests/reference/fb_1bpp.raw", "rb") as fp:
```
thijstriemstra, line 64: use pathlib.

rm-hull:
Framebuffer driver working on ILI9341 using  `/boot/config.txt` settings:

```
dtoverlay=pitft28-resistive,speed=64000000
```
Reset pin wired to GPIO 25
DC pin wired to GPIO 24
Backlight wired to  GPIO 2

(The setting for these pins appear to be hard wired in the kernel driver)

```console
$ dmesg | grep fb
[    1.004746] pci_bus 0000:00: root bus resource [mem 0x600000000-0x603ffffff] (bus address [0xf8000000-0xfbffffff])
[    4.951969] fbtft: module is from the staging directory, the quality is unknown, you have been warned.
[    4.955727] fb_ili9340: module is from the staging directory, the quality is unknown, you have been warned.
[    4.956503] fbtft_of_value: buswidth = 8
[    4.956523] fbtft_of_value: debug = 0
[    4.956539] fbtft_of_value: rotate = 90
[    4.956555] fbtft_of_value: fps = 25
[    5.140198] graphics fb0: fb_ili9340 frame buffer, 320x240, 150 KiB video memory, 4 KiB buffer memory, fps=25, spi0.0 at 64 MHz
```

image

The perfloop.py demo shows a frame rate of approx 14 frames/sec using the fb driver, verses approx 7 frames/sec using the ili9341 driver

luma/core/device.py
rm-hull: done

tox.ini
rm-hull: done

luma/core/device.py
rm-hull: refactored code

luma/core/cmdline.py
rm-hull: done

luma/core/device.py
rm-hull: tempted not to do that - these implementations are specific to how the RGB pixels are structured in the framebuffer.

the 1-bit framebuffer conversion is a bit weird, and i actually decided to take it out along with the 32-bit ARGB version on the basis that we can add them at a later date if there is a demand (unlikely)

luma/core/device.py
```diff
@@ -168,6 +173,90 @@ def display(self, image):
         :param image: Image to display.
         :type image: PIL.Image.Image
         """
-        assert(image.size == self.size)
+        assert image.size == self.size
 
         self.image = self.preprocess(image).copy()
+
+
+class fb(device):
```
thijstriemstra, line 181: im not sure if the class name should/can be 2 chars long.

rm-hull: Just following how the framebuffer is named in Linux (e.g. `/dev/fb0`)... but happy to change this to something like `luma.core.device.linux_frambuffer`. 

_[I didnt want to call this just `luma.core.device.framebuffer` as it would clash with the namespace of `luma.core.framebuffer`]_

thijstriemstra: `FrameBuffer`.

luma/core/device.py
```diff
@@ -168,6 +173,90 @@ def display(self, image):
...
+        given /dev/fbX framebuffer device.
+
+        :param image: Image to display.
+        :type image: PIL.Image.Image
+        """
+        assert image.mode == self.mode
+        assert image.size == self.size
+
+        image = self.preprocess(image)
+        path = "/dev/fb{}".format(self.id)
```
thijstriemstra, line 258: use f-string

tests/test_fb_device.py
```diff
@@ -0,0 +1,100 @@
...
+    ) as fake_open:
+        device = fb("/dev/fb1")
+        assert device.bits_per_pixel == 24
+        fake_open.assert_has_calls(
+            [call("/sys/class/graphics/fb1/bits_per_pixel", "r")]
+        )
+
+
+def test_display_16bpp():
+    with open("tests/reference/fb_16bpp.raw", "rb") as fp:
```
thijstriemstra, line 64: use helper reference file function for this

tests/test_fb_device.py
```diff
@@ -0,0 +1,100 @@
...
+            draw.rectangle((64, 0, 128, 32), fill="yellow")
+            draw.rectangle((0, 32, 64, 64), fill="orange")
+            draw.rectangle((64, 32, 128, 64), fill="white")
+
+    fake_open.assert_has_calls([call("/dev/fb1", "wb")])
+    fake_open.return_value.write.assert_called_once_with(reference)
+
+
+def test_display_24bpp():
+    with open("tests/reference/fb_24bpp.raw", "rb") as fp:
```
thijstriemstra, line 80: use helper reference file function for this

luma/core/device.py
rm-hull: fixed + updated in a few other places too

tests/test_fb_device.py
rm-hull: done

tests/test_fb_device.py
rm-hull: done

tests/test_fb_device.py
rm-hull: Uses get_reference_file instead (which uses pathlib)

thijstriemstra (approved):

PR #195, opened 2020-10-10
Remove cs_high capability

rm-hull:
This functionality was removed in the 5.4 kernel line, and this causes SPIDEV to fail, but the error manifests in luma.core when setting the cs_high line.

The specialist class `gpio_cs_spi` still supports setting the CS active high if necessary.

This issue has already been reported in several places, so this PR should address these:
* Fixes https://github.com/rm-hull/luma.led_matrix/issues/225
* Fixes https://github.com/rm-hull/luma.oled/issues/292
* Fixes https://github.com/rm-hull/luma.examples/issues/124


CHANGES.rst
```diff
@@ -4,6 +4,8 @@ ChangeLog
 +------------+---------------------------------------------------------------------+------------+
 | Version    | Description                                                         | Date       |
 +============+=====================================================================+============+
+| *TBC*      | * Remove SPI cs_high capability                                     |            |
```
thijstriemstra, line 7: maybe expand why it was removed.

thijstriemstra (dismissed):

rm-hull:
v1.17.2 released with this change

PR #198, opened 2020-10-13
tests: use pathlib

thijstriemstra:


rm-hull (dismissed):
Thanks 🙏 
I’m resisting temptation to use the framebuffer PR as a catch-all for minor snags

thijstriemstra:
PR still needs some work, will take a look at it later today.

rm-hull (approved):

PR #197, opened 2020-10-11
Drop support for Python 3.5

rm-hull:
Travis reports (see https://travis-ci.org/github/rm-hull/luma.core/jobs/734806373#L435):

```
DEPRECATION: Python 3.5 reached the end of its life on September 13th, 2020. Please upgrade
your Python as Python 3.5 is no longer maintained. pip 21.0 will drop support for Python 3.5
in January 2021. pip 21.0 will remove support for this functionality.
```

PR #199, opened 2020-10-16
tests: fix failure

thijstriemstra:


thijstriemstra:
@rm-hull any idea why none of the images match anymore? is it a different pillow version on travis that generates a diff image or something?

https://travis-ci.org/github/rm-hull/luma.core/jobs/736320999

thijstriemstra:
The tests for py35 are failing (it uses an older version of Pillow because they dropped support for py35): https://travis-ci.org/github/rm-hull/luma.core/jobs/736337071

But this should go away when #197 is merged first.

rm-hull (approved):

PR #202, opened 2020-10-18
Pin Pillow to last version before 8.0.0

rm-hull:


thijstriemstra:
What is the rationale for this change? Pillow 8 produces better results.

thijstriemstra (changes requested):
What is the rationale for this change? Pillow 8 produces better results. I'm fine with pinning something like pytest to an older version, like we do now, but Pillow is main dependency and we should always try to support the latest and greatest version, especially when it's producing better results.

rm-hull:
_Not required_

PR #205, opened 2020-10-24
Improved diff-to-previous framebuffer performance

rm-hull:
NOTE: This will require a 2.0.0 release since it changes the interface used by the `full_frame` and `diff_to_previous` classes. These classes are (or should only) be used by **luma.lcd** and **luma.oled** packages so there are two PRs in those respective repos to pin them to the 1.x.y line. Once PR is merged & released, those repos will be updated to use the new interface.

### What does this PR do?
Prior to this PR, the diff-to-previous algorithm would just do a diff across the whole image vs. the previous image, and get the bounding box of the change. If there were lots of changes across the screen (e.g the bouncing balls or jet-set willy demo) then the bounding box would be more-or-less the whole screen.

This change will split the screen into segments, and do a diff per segment, so for example: if it were split into 25 segments (a 5x5 grid) and there is a change in the top-left corner and another change in the bottom-right corner, then two small changesets would result. (With the current diff-to-previous, this would generate one changeset for the entire screen). This should realise a sizeable performance improvement for partial screen updates, animations and scrolling artefacts, especially on larger displays (e.g. ILI9341)

All the devices in child repos that use the framebuffer will need to be updated to handle multiple changesets, but this should be negligible.


luma/core/framebuffer.py
```diff
@@ -89,33 +96,18 @@ class full_frame(object):
     :param device: The target device, used to determine the bounding box.
     :type device: luma.core.device.device
     """
-    def __init__(self, device):
-        self.bounding_box = (0, 0, device.width, device.height)
 
-    def redraw_required(self, image):
+    def __init__(self, **kwargs):
```
thijstriemstra, line 100: is this necessary to override?

luma/core/mixin.py
```diff
@@ -10,6 +10,7 @@ class capabilities(object):
     This class should be 'mixed-in' to any :py:class:`luma.core.device.device`
     display implementation that should have "device-like" capabilities.
     """
+
```
thijstriemstra, line 13: not needed.

luma/core/framebuffer.py
rm-hull: I think the implementations will call this dynamically passing the device like here: https://github.com/rm-hull/luma.oled/blob/master/luma/oled/device/color.py#L25, and I may want to pass in the num_segments too, so we can either use `**kwargs` or have `device, num_segments` like the diff-to-previous constructor

CHANGES.rst
```diff
@@ -4,6 +4,8 @@ ChangeLog
 +------------+---------------------------------------------------------------------+------------+
 | Version    | Description                                                         | Date       |
 +============+=====================================================================+============+
+| **2.0.0**  | * BREAKING CHANGE: Improved framebuffer performance                 | TBC        |
```
thijstriemstra, line 7: I'm not a fan of starting with the "negative" and then pointing out a big positive. Adding `(breaking change)` lowercase at end maybe? Same for the title of this pull request.

luma/core/mixin.py
rm-hull: done

CHANGES.rst
rm-hull: sure thing

rm-hull: although, this 'breaking change' would not be visible to consumers of the displays, only between luma.core <=> luma.lcd, luma.oled

luma/core/framebuffer.py
thijstriemstra: in that case, maybe add a docstring so you can omit the `pass` statement altogether.

thijstriemstra:
> Once PR is merged & released, those repos will be updated to use the new interface.

I'd really like to see PRs in sub-projects that use this change without having to release luma.core 2.0.0 first. So I think that would mean temporarily switching luma.core installation to git version of luma.core (particular 2.0.0 branch maybe) in a sub-project and test it against that. Otherwise I wouldn't be suprised to see many 2.0.x releases and even more backward incompatible changes.

CHANGES.rst
thijstriemstra: yea, downplaying it isn't a problem, this looks better.

rm-hull:
Yeah, my intention is to to `pip install -e .` locally for this branch and likewise for the sub-project. If it works for one device in luma.lcd, then it will work for all devices across luma.lcd and luma.oled. 

I wont mark this as ready or release until it is QA'd on a real device along with evidence of better performance

luma/core/framebuffer.py
rm-hull: Refactored API so constructor isn't needed at all

rm-hull:
@thijstriemstra I think this is pretty much stable now; I don'i intend to make any further changes here. There will be a series of PRs in the oled and lcd packages to make use of this new class, but they are still draft ATM. Initial impressions are very positive: there is a HUGE performance gain on the ILI9341 driver implementation, but I still have some test failures to investigate/squash first, see https://github.com/rm-hull/luma.lcd/pull/129

thijstriemstra:
Sounds good. Request review when ready.

luma/core/framebuffer.py
```diff
@@ -6,15 +6,17 @@
 Different implementation strategies for framebuffering
 """
 
-from PIL import Image, ImageChops
+import math
```
thijstriemstra, line 9: can you `from math import sqrt` instead

luma/core/framebuffer.py
```diff
@@ -6,15 +6,17 @@
 Different implementation strategies for framebuffering
```
thijstriemstra, line 6: copyright date in header needs to be updated.

luma/core/framebuffer.py
```diff
@@ -25,57 +27,71 @@ class diff_to_previous(object):
...
+        segment the image entirely in both width and height. i.e setting to 9 will
+        subdivide the image into a 3x3 grid when comparing to the previous image.
+    :type num_segments: int
     """
-    def __init__(self, device):
-        self.image = Image.new(device.mode, device.size, "white")
-        self.bounding_box = None
 
-    def redraw_required(self, image):
+    def __init__(self, num_segments=4, debug=False):
```
thijstriemstra, line 37: debug param needs to be documented.

luma/core/framebuffer.py
```diff
@@ -25,57 +27,71 @@ class diff_to_previous(object):
...
+        self.prev_image = None
+
+    def redraw(self, image):
         """
-        Calculates the difference from the previous image, return a boolean
-        indicating whether a redraw is required. A side effect is that
-        ``bounding_box`` and ``image`` attributes are updated accordingly, as is
-        priming :py:func:`getdata`.
+        Calculates the difference from the previous image, returning a sequence of
+        image sections and bounding boxes that changed since the previous image.
```
thijstriemstra, line 46: this sentence probably needs another linebreak so the note is separated by a newline.

luma/core/framebuffer.py
```diff
@@ -89,33 +105,15 @@ class full_frame(object):
     :param device: The target device, used to determine the bounding box.
     :type device: luma.core.device.device
```
thijstriemstra, line 106: this device type probably needs `:py:class:` with backticks

luma/core/framebuffer.py
```diff
@@ -89,33 +105,15 @@ class full_frame(object):
...
     """
-    def __init__(self, device):
-        self.bounding_box = (0, 0, device.width, device.height)
 
-    def redraw_required(self, image):
+    def redraw(self, image):
         """
         Caches the image ready for getting the sequence of pixel data with
```
thijstriemstra, line 111: this docstring mentions it uses `getdata` but does it still do that?

CHANGES.rst
```diff
@@ -4,7 +4,8 @@ ChangeLog
 +------------+---------------------------------------------------------------------+------------+
 | Version    | Description                                                         | Date       |
 +============+=====================================================================+============+
-| *TBC*      | * Allow a wider range of SPI bus speeds                             |            |
+| **2.0.0**  | * Improved framebuffer performance (breaking change)                | TBC        |
```
thijstriemstra, line 7: do we have an upgrade document where breaking changes can be described in more detail?

thijstriemstra (changes requested):

luma/core/framebuffer.py
rm-hull: device is no longer used, so removed

luma/core/framebuffer.py
rm-hull: Reworded docs

luma/core/framebuffer.py
rm-hull: Done

luma/core/framebuffer.py
rm-hull: done

luma/core/framebuffer.py
rm-hull: done

luma/core/framebuffer.py
rm-hull: done

thijstriemstra (approved):

CHANGES.rst
rm-hull: Not really ... this is really an internal change to the implementation of luma.core and its dependent projects, rather than a material change to the public interface.

I have added things like the following to the "user-facing" API docs: https://github.com/rm-hull/luma.oled/pull/310/files#diff-b235326d22f35fd32f51120f8cccb66f7440e2f26997be86709b3afb8e29ce8aR343 and if there are specific instances where someone explicitly passing a string of "full_frame" or "diff_to_previous" it will show a runtime warning, e.g.: https://github.com/rm-hull/luma.oled/pull/310/files#diff-b235326d22f35fd32f51120f8cccb66f7440e2f26997be86709b3afb8e29ce8aR841-R849

thijstriemstra: > this is really an internal change to the implementation of luma.core 

I would stop calling it a breaking change in that case. maybe bckwards incompatible or just omit it?

Issue #204, opened 2020-10-20
Use time.perf_counter_ns


thijstriemstra:
Python 3.7 introduces `time.monotonic_ns`. [PEP564](https://www.python.org/dev/peps/pep-0564/) explains the rationale but what stood out to me is:

> A server is running for longer than 104 days. A clock is read before and after running a function to measure its performance to detect performance issues at runtime. Such benchmark only loses precision because of the float type used by clocks, not because of the clock resolution.

> On Python microbenchmarks, it is common to see function calls taking less than 100 ns. A difference of a few nanoseconds might become significant.

More precision is always better.

thijstriemstra:
thoughts @rm-hull? It could be introduced in 2.0 as well.

rm-hull:
Yep, we can add this; was there anywhere specifically it would be useful - presumably updating existing instances of `time.monotic()` only?

thijstriemstra:
yea my thoughts as well. I'll start a branch.

thijstriemstra:
This can be easily supported once Python 3.6 reaches EOL.

thijstriemstra:
> This can be easily supported once Python 3.6 reaches EOL.

looks like this happened on [23 Dec 2021](https://endoflife.date/python)

thijstriemstra:
more good `time.sleep()` improvements in python 3.11: https://docs.python.org/3.11/whatsnew/3.11.html#time

PR #206, opened 2020-10-25
use time.perf_counter

thijstriemstra:
refs #204


rm-hull (dismissed):

thijstriemstra:
@rm-hull I changed the calls to `perf_counter_ns`. `perf_counter_ns` seems to be a more suitable api for what we're trying to use it (available since python 3.3 so that explains why we didn't use it earlier). The tests do not really account for nano or milliseconds so they still pass. Question is, should we change the APIs and expect nanoseconds, where it expected seconds before, or should the api stay the same and we will do the nanosecond/second conversion for the user?

thijstriemstra:
@rm-hull I wanted to include this and a few other backward incompatible changes in v2.0. Isn't v2.0 currently breaking all other sub-projects? I'm confused.

rm-hull:
Sorry the notification for your last msg ended up in spam. The framebuffer changes would've only affected certain LCD and OLED devices.

rm-hull:
These changes shouldn't be breaking should they? We are just using a different internal mechanism to record time more accurately. The interface still stays the same doesn't it?

rm-hull (commented):
Time.perf_counter_ns() returns an integer which is in nanoseconds according to https://docs.python.org/3/library/time.html#time.perf_counter_ns.

So the elapsed time will be in nanoseconds, which if called the sleep at line 207 would be a very long time. If we are to use this function, tit should divide by 1e9 to get the correct number of seconds presumably?

... or just use time.perf_counter() ?

rm-hull:
And if we went with perf_counter(), then it would work with python 3.6 too

thijstriemstra:
But perf_counter doesn't have the additional precision aka it'll start drifting after only 104 days.

thijstriemstra:
> But perf_counter doesn't have the additional precision aka it'll start drifting after only 104 days.

Let's start with `perf_counter` and improve with more precision later.

rm-hull (approved):

PR #208, opened 2020-10-26
travis: test on ubuntu focal

thijstriemstra:


rm-hull (dismissed):

rm-hull:
huh weird ... 3 hours later and no builds have gone through: Is travis having issues or is it because the PR is marked as draft ? 

in any regard, i've noticed that builds are taking a long time to start on travis these days...

thijstriemstra:
the builds did start: https://travis-ci.org/github/rm-hull/luma.core/jobs/738983807

but some weird virtualenv error..

rm-hull:
See also https://travis-ci.community/t/build-delays-for-open-source-project/10272/9

thijstriemstra:
We're currently modifying the travis system python packages by upgrading pip, setuptools, tox and travis libraries.

It's probably better to create a separate virtualenv and upgrade everything there. Tox will then create separate virtualenvs during it's testing. This way we don't have to fiddle with the system and potentially break something.

.travis.yml
```diff
@@ -15,6 +15,8 @@ matrix:
         - python: 3.8
           env: TOXENV=qa,doc
 
-install: pip install --ignore-installed --upgrade setuptools pip tox coveralls
+install:
+  - pip uninstall virtualenv
```
rm-hull, line 19: see: https://travis-ci.org/github/rm-hull/luma.core/jobs/740002770#L202-L207

Add a `-y` flag maybe ?

thijstriemstra: ill rework this pr this weekend. that last commit was a (dumb) test.

rm-hull (approved):

PR #218, opened 2020-11-27
ci: switch to github actions

### LINKED ISSUE #216: Reduce the build jobs on Travis ... or use Github Actions
Email today from travis:

> Welcome to Travis CI, Richard Hull!You are on Free.You have 10,000 credits left - these will begin counting down automatically as soon as you run your first build.You can use your credits to build on both private and open-source repositories using Linux, macOS, and Windows OS.Use the `Sign up now` button below to sign up for a bigger plan to ensure the builds for your private repositories continue uninterrupted.If you're building open-source public repositories and want to signup for an open-source plan, please contact Travis CI.
--
You have 10,000 Credits left for building on Travis CI!

Now that travis is no longer 'free', and since every time a build across any luma.* project is triggered, it will kick off multiple jobs to run the same tests on different versions of python.

This will quickly drain down the 10000 credits, so we have a choice:

1. Remove the build jobs for different versions of python (this reduces the rate at which the credits will go down, but not ideal in the longer term)
1. Contact travis and appeal for an open-source plan (this may be successful, but who knows?)
1. Move all CI operations to github actions instead (free right now, and some small effort required to convert current CI)

Any thoughts on preferred next step @thijstriemstra  ?

thijstriemstra:
fixes #216


thijstriemstra:
haven't found a solution for coveralls yet but other than that it should be ready.

rm-hull (approved):
That’s fantastic. And while does the QA check on each version, it looks to be quicker than Travis.

Re. The coveralls stuff, for some of my other non python projects I seem to remember just doing a curl to post the results to a coveralls endpoint: https://github.com/rm-hull/lein-nvd/blob/master/.travis.yml#L17 for example. Presumably we could just do something similar at some point

PR #219, opened 2020-11-29
Linux framebuffer to use diff-to-previous algorithm

### LINKED ISSUE #217: Possible to use diff-algorithm with linux_framebuffer device?
#### Type of Raspberry Pi

Raspberry Pi 3 Model B (v1.2)

#### Linux Kernel version

`Linux raspberrypi 5.4.51-v7+ #1333 SMP Mon Aug 10 16:45:19 BST 2020 armv7l GNU/Linux`
(Buster release of Raspberry Pi OS)

#### Question

Thanks for adding the linux_framebuffer device into luma.core!  I'm just getting the chance to experiment with it.

Is it possible at all to make use of luma's **diff_to_previous** algorithm for updates to the fbdev framebuffer?  I ask just since the full update (I assume that's what is currently being used) is taking a little over 1/4 of a second.

With an update loop that looks as follows:
```
        while True:
            try:
                start_time = datetime.now()
                draw.rectangle([(0,0), (frame_size[0],frame_size[1])], 'black', 'black')
                draw.text(( 5, 5), "Hello, world",  fill='white', font=font_main)
                device.display(image)
                #update_display()   -- isolate update time to just fb update
                print(datetime.now(), "update_display() took ", datetime.now() - start_time)
```

some representative results are
```
2020-11-27 11:27:25.556874 update_display() took  0:00:00.292568
2020-11-27 11:27:26.745018 update_display() took  0:00:00.276966
2020-11-27 11:27:27.929180 update_display() took  0:00:00.273018
2020-11-27 11:27:29.115797 update_display() took  0:00:00.275477
2020-11-27 11:27:30.320051 update_display() took  0:00:00.293113
2020-11-27 11:27:31.510489 update_display() took  0:00:00.279289
2020-11-27 11:27:32.691867 update_display() took  0:00:00.270262
2020-11-27 11:27:33.900963 update_display() took  0:00:00.297949
2020-11-27 11:27:35.108634 update_display() took  0:00:00.296486
2020-11-27 11:27:36.315688 update_display() took  0:00:00.295907
2020-11-27 11:27:37.499120 update_display() took  0:00:00.272288
2020-11-27 11:27:38.698876 update_display() took  0:00:00.288624
2020-11-27 11:27:39.891200 update_display() took  0:00:00.281173
2020-11-27 11:27:41.077159 update_display() took  0:00:00.274793
```

The above is with an ILI9341-based display driven using tinyDRM's `mi0283qt` driver.  The dtoverlay file is currently specifying a SPI frequency of 32 MHz.  The SPI bus could certainly be set faster, but I thought I'd inquire about the fbdev update.

Thanks again!
  Matt

rm-hull:
Doing a full-frame render, I observe 14 frames/sec on RPi 4 using the bounce.py demo on the linux framebuffer pseudo-device.

With this PR, where it uses the diff-to-previous render algorithm, the same bounce demo shows the frame rate at approx 30 frames/sec

Fixes #217


thijstriemstra:
@mattblovell can you also take a look at this PR?

luma/core/device.py
```diff
@@ -245,7 +245,7 @@ def __toRGB565(self, image):
             yield r & 0xF8 | g >> 5
 
     def __toRGB(self, image):
-        return image.tobytes()
+        return iter(image.tobytes())
```
thijstriemstra, line 248: this looks breaking so 2.1.0 sounds like a good version nr.

luma/core/device.py
```diff
@@ -3,11 +3,14 @@
 # See LICENSE.rst for details.
 
 import os
+import mmap
```
thijstriemstra, line 6: from mmap import mmap?

luma/core/device.py
```diff
@@ -250,13 +254,31 @@ def display(self, image):
...
-        data = bytes(self.__image_converter(image))
 
-        with open(path, "wb") as fp:
-            fp.write(data)
+        bytes_per_pixel = self.bits_per_pixel // 8
+        image_bytes_per_row = self.width * bytes_per_pixel
+
+        with open(path, "w+b") as fp:
+            with mmap.mmap(fp.fileno(), self.width * self.height * bytes_per_pixel,
+                           flags=mmap.MAP_SHARED, access=mmap.PROT_WRITE) as mm:
```
mattblovell, line 271: Could the fp and memory-mapping be performed once?  I don't know if the expected sequence of calls make for any other good opportunity or if display() could just perform that work once, holding on to the file pointer and mapping for all subsequent calls?

Not that more speed is necessarily called for.

mattblovell (dismissed):
Left one comment wondering if the fbdev file open and memory mapping could be performed once, holding on to the results for the duration of program execution.

Update speed (in my dummy loop that doesn't actually update anything) has definitely improved.  I'll go run some experiments with a non-trivial update loop, but seems good to go from my limited perspective.

luma/core/device.py
rm-hull: I did try that initially, and noticed that the frame rate was initially lower (25 FPS vs. 30 FPS) and observed it degenerate over time (tending towards 20 FPS) over several runs of the bounce.py demo. I don’t know why this would be, as one might expect it to be marginally quicker. I did wonder if it was down to using instance variables rather than local variables (unlikely) but I have a feeling it down to mmap performance somehow (https://stackoverflow.com/a/38682857) or my overuse of yield/generators. I might try and rework it again to just use file `seek`ing instead to eliminate whether mmap is the culprit.

Another side effect of initialising fp and mm once is that there would be more tests that would need to be repaired, but I guess that is just me being lazy...

rm-hull: Ok, so seeking and writing to the file pointer directly (rather than going through mmap), and the frame rate jump to 68 FPS, so I think we can saffely abandon mmap altogether

rm-hull: Keeping the file handle open: frame rate goes to 73 FPS

luma/core/device.py
```diff
@@ -192,11 +194,13 @@ class linux_framebuffer(device):
         is given, the device is determined from the `FRAMEBUFFER` environmental
         variable instead. See https://www.kernel.org/doc/html/latest/fb/framebuffer.html
         for more details.
+    :param framebuffer: Framebuffering strategy, currently instances of
+        ``diff_to_previous`` or ``full_frame`` are only supported.
```
thijstriemstra, line 198: mention it uses diff_to_previous by default.

luma/core/device.py
```diff
@@ -208,8 +212,12 @@ def __init__(self, device=None, **kwargs):
         assert self.bits_per_pixel in image_converters, f"Unsupported bit-depth: {self.bits_per_pixel}"
         self.__image_converter = image_converters[self.bits_per_pixel]
 
+        self.framebuffer = framebuffer or diff_to_previous(num_segments=16)
         self.capabilities(width, height, rotate=0, mode="RGB")
 
+        path = f"/dev/fb{self.id}"
+        self.__fp = open(path, "wb")
```
thijstriemstra, line 219: even though it's private, I'd prefer a more descriptive name for fp

rm-hull: fixed

luma/core/device.py
rm-hull: fixed

luma/core/device.py
```diff
@@ -255,8 +267,20 @@ def display(self, image):
         assert image.size == self.size
 
         image = self.preprocess(image)
-        path = f"/dev/fb{self.id}"
-        data = bytes(self.__image_converter(image))
-
-        with open(path, "wb") as fp:
-            fp.write(data)
+        file_handle = self.__file_handle
```
thijstriemstra, line 270: is this ref necessary?

rm-hull: primarily: cheaper to reference local variable than repeatedly lookup of self.variable in loop a bit later on in the code, but TBH probably a micro-optimization

thijstriemstra (approved):

Issue #222, opened 2020-12-06
Supporting 32-bit depth for linux_framebuffer


mattblovell:
#### Type of Raspberry Pi

Raspberry Pi Zero W

#### Linux Kernel version

Linux raspberrypi 5.4.79+ #1373 Mon Nov 23 13:18:15 GMT 2020 armv6l GNU/Linux

(Dec 2020 update of buster for Raspberry Pi OS)

#### Enhancement question

I'm trying out a small HDMI-attached display that I had working previously under an RPi 4.  For some reason, on the Zero, the framebuffer device gets created with 32-bit depth.  As a consequence, luma.core doesn't want to start.  Invoking one of the examples like

`python3 demo.py --display linux_framebuffer --framebuffer-device /dev/fb0`

yields

```
  File "/home/pi/projects/luma.core/luma/core/device.py", line 212, in __init__
    assert self.bits_per_pixel in image_converters, f"Unsupported bit-depth: {self.bits_per_pixel}"
AssertionError: Unsupported bit-depth: 32
```

Unlike the RPi4, I think the Zero continues to use legacy driver for its display.  With that, one can specify

`framebuffer_depth=24`

in `/boot/config.txt`.  Using that, and rebooting, I am able to get the examples to start, but there is an odd color-shift...

- red appears blue
- blue is red
- orange is light blue
- yellow is sky blue
- green is still green

Adding `--bgr` to the command line doesn't change the results.

rm-hull:
Couple of things to unpack here:
* I didn’t have a 32 bit depth device to test with so chose not to provide an implementation (it did not occur to me at the time to plug in a HDMI display, which seems obvious now)
* It does sound like flag needs to be supplied to flip red & blue for 24 bit displays. I think we just assumed pixels would be RGB, rather than BGR. Supporting BGR may well incur a performance penalty though in that mode.

mattblovell:
> * It does sound like flag needs to be supplied to flip red & blue for 24 bit displays. I think we just assumed pixels would be RGB, rather than BGR.

I was surprised as well, since I'm pretty sure I tried out that HDMI display on a Pi4 with luma.lcd and linux_framebuffer without issue.  (That was the setup that triggered my question about the diff algorithm for fbdev.)  I may not have been using the legacy driver with that setup, though.

Perhaps it's an artifact of using `config.txt` to limit color depth.  The Raspberry Pi page

https://www.raspberrypi.org/documentation/configuration/config-txt/video.md

[...]

PR #223, opened 2020-12-08
linux framebuffer: Support BGR, RGBA & BGRA colorspace displays

rm-hull:
Refs #222


thijstriemstra:
> Adding --bgr to the command line doesn't change the results.

Do we need a new cmdline option for this?

rm-hull:
it should pass it in automatically by virtue of https://github.com/rm-hull/luma.core/blob/master/luma/core/cmdline.py#L237-L238 now that the class can pluck out the value.

@mattblovell could you try checking out this branch, installing it on your Pi Zero and let us know if this corrects the colours (please add the `--bgr` flag)? If it does, I will add a regression test and bundle a new release.

rm-hull:
also... assuming it does work, i wonder if there is a way to sense from anything in `/sys/class/graphics/fbX/*` whether the device is RGB or BGR ?

mattblovell:
Pulling top of `bugfix/bgr-flag-for-linux-framebuffer` and adding `--bgr` to the command line:

`python3 demo.py --bgr --display linux_framebuffer --framebuffer-device /dev/fb0`

is getting an error.  Here is the complete output:

```
Version: luma.core 2.1.0 (luma.core 2.1.0)
Display: linux_framebuffer
Interface: i2c
Dimensions: 800 x 480
------------------------------------------------------------
Testing basic canvas graphics...
Traceback (most recent call last):
  File "demo.py", line 99, in <module>
    main()
  File "demo.py", line 59, in main
    primitives(device, draw)
  File "/home/pi/projects/luma.core/luma/core/render.py", line 43, in __exit__
    self.device.display(self.image)
  File "/home/pi/projects/luma.core/luma/core/device.py", line 282, in display
    generator = self.__image_converter(image)
  File "/home/pi/projects/luma.core/luma/core/device.py", line 255, in __toBGR
    return iter(image[:, :, [2, 1, 0]].tobytes())
TypeError: 'Image' object is not subscriptable
```

The `demo.py` invocation still works as it did before (colors swapped) leaving off the `--bgr` option.

mattblovell:
> ... assuming it does work, i wonder if there is a way to sense from anything in /sys/class/graphics/fbX/* whether the device is RGB or BGR ?

Here's are the available files on the Pi Zero:
```
pi@raspberrypi:~/projects/luma.core $ ll /sys/class/graphics/fb0/
total 0
-rw-r--r-- 1 root root 4096 Dec  8 18:22 bits_per_pixel
-rw-r--r-- 1 root root 4096 Dec  8 18:27 blank
-rw-r--r-- 1 root root 4096 Dec  8 18:27 bl_curve
-rw-r--r-- 1 root root 4096 Dec  8 18:27 console
-rw-r--r-- 1 root root 4096 Dec  8 18:27 cursor
-r--r--r-- 1 root root 4096 Dec  8 18:27 dev
lrwxrwxrwx 1 root root    0 Dec  8 18:27 device -> ../../../soc:fb
-rw-r--r-- 1 root root 4096 Dec  8 18:27 mode
-rw-r--r-- 1 root root 4096 Dec  8 18:27 modes
-r--r--r-- 1 root root 4096 Dec  7 00:04 name
-rw-r--r-- 1 root root 4096 Dec  8 18:27 pan
drwxr-xr-x 2 root root    0 Dec  8 18:27 power
-rw-r--r-- 1 root root 4096 Dec  8 18:27 rotate
-rw-r--r-- 1 root root 4096 Dec  8 18:27 state
-r--r--r-- 1 root root 4096 Dec  8 18:27 stride
lrwxrwxrwx 1 root root    0 Dec  7 00:03 subsystem -> ../../../../../../class/graphics
-rw-r--r-- 1 root root 4096 Dec  7 00:03 uevent
-rw-r--r-- 1 root root 4096 Dec  8 18:22 virtual_size
```

`bits_per_pixel`, with the override present in `/boot/config.txt` contains `24`.
`stride` has `2400`
`modes` has `U:800x480p-0`
`virtual_size` has `800,480`

This RPi forum thread

https://www.raspberrypi.org/forums/viewtopic.php?t=257989

discusses endian-ness, but doesn't touch on any means of discovering color order short of "fill some fbdev bytes and look".

mattblovell:
Oddly, with the 24-bit depth override still present, this little program

https://github.com/jumpnow/fbtest

still behaves normally.  Invoking

` ./fbtest -r0xff -g0 -b0x0`

gets one a screen entirely filled with red, and 

` ./fbtest -r0 -g0 -b0xff`

gets one a blue screen.  

I'll try removing the `/boot/config.txt` directive and see what I end up with.

mattblovell:
The `fbtest` program ends up using ioctl queries to get fbdev information.  See FBIOGET_FSCREENINFO and FBIOGET_VSCREENINFO discussion in Section 3 of

https://www.kernel.org/doc/Documentation/fb/api.txt

BTW, I feel somewhat guilty encouraging work on fbdev support , since Linux appears to be deprecating it.  It stills seems like an appropriate fit for little SPI or DSI displays, but Linux as a whole is focused on much fancier GPUs and thus DRM/KMS.  I've had short (and friendly) email exchanges with two Linux kernel developers regarding how the current desktop focus raises the "barrier of entry" for small, 2D, non-accelerated graphics applications.  Maybe that will do some good (some day).

Nevertheless, I just wanted to ensure you were aware of the above.

rm-hull:
> since Linux appears to be deprecating it

Yep. I have a feeling that it has been deprecating for quite a number of years, but guess we should probably not spend too much time on this PR.

In the meantime could you try the latest wip commit i made on the branch. It is untested but should at least not crash this time 😬

mattblovell:
> In the meantime could you try the latest wip commit i made on the branch. It is untested but should at least not crash this time

Not only does it not crash, but blue is blue and red is red (when using `--bgr`)! 👍

rm-hull:
And I just added another commit to support 32 bits-per-pixel (RGBA colour depth) - again untested, so if you could perhaps try it out please

mattblovell:
> And I just added another commit to support 32 bits-per-pixel (RGBA colour depth) - again untested, so if you could perhaps try it out please

That is working (I verified that `/sys/class/graphics/fb0/bits_per_pixel` is returning 32).

However, the blue/red mixup is back, and `--bgr` doesn't seem to have any effect.

rm-hull:
Erg, so it is actually BGRA... so watch out for another converter coming shortly

mattblovell:
BTW, here is what `fbtest`'s output has to say regarding the two color depth modes, in case it's interesting:

**32-bit depth**

```
======= FB VAR SCREENINFO =======
xres: 800
yres: 480
yres_virtual: 480
buffer number: 1
bpp : 32
red bits    :
    offset   : 16
    length   : 8
    msb_right: 0
green bits  :
    offset   : 8
    length   : 8
    msb_right: 0
blue bits   :
    offset   : 0
    length   : 8
    msb_right: 0
transp bits :
    offset   : 24
    length   : 8
    msb_right: 0
=================================
======= FB FIX SCREENINFO =======
id          : BCM2708 FB
smem_start  : 0x1E90C000
smem_len    : 1536000
line_length : 3200
=================================
```

**24-bit depth**  (after updating `/boot/config.txt` and a reboot)

```
======= FB VAR SCREENINFO =======
xres: 800
yres: 480
yres_virtual: 480
buffer number: 1
bpp : 24
red bits    :
    offset   : 16
    length   : 8
    msb_right: 0
green bits  :
    offset   : 8
    length   : 8
    msb_right: 0
blue bits   :
    offset   : 0
    length   : 8
    msb_right: 0
transp bits :
    offset   : 24
    length   : 0     <----
    msb_right: 0 
=================================
======= FB FIX SCREENINFO =======
id          : BCM2708 FB
smem_start  : 0x1E96A000
smem_len    : 1152000
line_length : 2400
=================================
```

The only difference, besides the bpp, is the 0 length noted for transp(arency) bits when in 24-bit mode.  Both seem to have blue in the least-significant byte (offset 0).  So, it is indeed odd that any color order swap seems necessary.  I'm confused anyway!

rm-hull:
alright, another commit - please retry 32-bit with `--bgr`

mattblovell:
> please retry 32-bit with `--bgr`

All is working!

I fear a display of this resolution is bit too much for Python running on a Pi Zero, unfortunately.  Poor `bounce.py` is only getting 5.7 fps. 

I'll have to try connecting the display to the RPi 2 and see how it does.  (Since I cannot make use of RPi4's dual HDMI connectors to actually have independent displays with the newer GPU drivers, I've been playing with how else to make use of the 4" HDMI panel.)

mattblovell:
> I'll have to try connecting the display to the RPi 2 and see how it does. 

Up and running with the top of this bugfix branch.  The `bounce.py` demo is achieving 15 to 15.4 fps at 480x800.

mattblovell:
Just as an aside, I may have too many displays... :)  

IMG-0717_sm

luma/core/device.py
```diff
@@ -188,35 +189,38 @@ class linux_framebuffer(device):
...
     .. versionadded:: 2.0.0
     """
 
-    def __init__(self, device=None, framebuffer=None, **kwargs):
+    def __init__(self, device=None, framebuffer=None, bgr=False, **kwargs):
         super(linux_framebuffer, self).__init__(serial_interface=noop())
         self.id = self.__get_display_id(device)
         (width, height) = self.__config("virtual_size")
         self.bits_per_pixel = next(self.__config("bits_per_pixel"))
         image_converters = {
```
thijstriemstra, line 211: a simple comment describing what image converters maps would be nice

luma/core/device.py
```diff
@@ -188,35 +189,38 @@ class linux_framebuffer(device):
...
-        self.__image_converter = image_converters[self.bits_per_pixel]
+        assert (self.bits_per_pixel, bgr) in image_converters, f"Unsupported bit-depth: {self.bits_per_pixel}"
+        self.__image_converter = image_converters[(self.bits_per_pixel, bgr)]
 
         self.framebuffer = framebuffer or diff_to_previous(num_segments=16)
         self.capabilities(width, height, rotate=0, mode="RGB")
-
-        path = f"/dev/fb{self.id}"
-        self.__file_handle = open(path, "wb")
+        self.__file_handle = open(f"/dev/fb{self.id}", "wb")
```
thijstriemstra, line 223: an explicit comment somewhere that this handle doesn't close until `cleanup`

luma/core/device.py
```diff
@@ -251,6 +255,17 @@ def __toRGB565(self, image):
     def __toRGB(self, image):
         return iter(image.tobytes())
 
+    def __toBGR(self, image):
```
thijstriemstra, line 258: i wouldn't mind moving all these helpers to luma.core.util

rm-hull: I'm just a bit reluctant to put them in there, because they aren't really intended to be general purpose: 
* The input image is always expected to be an RGB image - making them for public consumption, we'd have to document that the images should be RGB only and/or add an `assert image.mode == 'RGB'` which i don't really want to do in a hot loop
* The return value is an iterator of bytes which is specifically chosen so that the looping around `islice()` is uncomplicated (see https://github.com/rm-hull/luma.core/blob/master/luma/core/device.py#L284). 
* There is nowhere else in the codebase that could readily use these methods AFAICS.

luma/core/device.py
rm-hull: fixed

luma/core/device.py
rm-hull: fixed

thijstriemstra (approved):

PR #228, opened 2021-01-06
Additional fonts

rm-hull:
Fixes #226


luma/core/legacy/font.py
```diff
@@ -22,6 +25,9 @@
 |                   | Spectrum.  Only contains characters 0x20-0x7E   |                                                                                                |
 |                   | inclusive; all others will appear as blanks.    |                                                                                                |
 +-------------------+-------------------------------------------------+------------------------------------------------------------------------------------------------+
+| ``SPECCY_FONT``   | Another font based on the character set from    | http://www.kreativekorp.com/software/fonts/index.shtml                                         |
```
thijstriemstra, line 28: use https instead

luma/core/legacy/font.py
```diff
@@ -8,6 +8,9 @@
 +-------------------+-------------------------------------------------+------------------------------------------------------------------------------------------------+
 | Font name         | Notes                                           | Source                                                                                         |
 +===================+=================================================+================================================================================================+
+| ``ATARI_FONT``    | Designed for a terminal emulator written for    | http://gnu.ethz.ch/linuks.mine.nu/atari/                                                       |
```
thijstriemstra, line 11: this link doesn't resolve. But better to have something that nothing..?

luma/core/legacy/font.py
```diff
@@ -1665,5 +1673,533 @@ def __getitem__(self, ascii_code):
     [0x00, 0x00, 0x00, 0x00],  # 0xFF
 ]
 
+#: Bit patterns for SPECCY_FONT
+#: (another font based on the character set from the ZX Spectrum).
+#: Source: http://www.kreativekorp.com/software/fonts/index.shtml,
```
thijstriemstra, line 1678: use https instead

thijstriemstra (changes requested):
See inline comments. I wouldn't use github username in pull request/commit message, as to not make an exception, or we'd have to do it for every contribution IMO.

luma/core/legacy/font.py
rm-hull: done

luma/core/legacy/font.py
rm-hull: yep, lets leave the link in just in case it ever came back

luma/core/legacy/font.py
rm-hull: done

thijstriemstra (dismissed):

thijstriemstra:
Ah, the link is broken, and detected by Sphinx: https://github.com/rm-hull/luma.core/runs/1657924751#step:8:192

I'll take a look.

thijstriemstra (changes requested):
Fixed it but table is malformed now: https://github.com/rm-hull/luma.core/runs/1658072788#step:8:247

thijstriemstra (approved):

PR #232, opened 2021-01-16
Use github actions to publish to pypi

### LINKED ISSUE #224: Use github actions to publish to pypi
See https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries

thijstriemstra:
fixes #224 

depends on #233


thijstriemstra:
@rm-hull It should be working, but how are we going to safely test this..

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,42 @@
...
+    - name: Build a binary wheel and a source tarball
+      run: >-
+        python -m
+        build
+        --sdist
+        --wheel
+        --outdir dist/
+        .
+    - name: Publish package to test PyPI
+      if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
```
rm-hull, line 30: Maybe changed this line to check for the presence of the word `test` in the tag name and line 37 to exclude it, so it would only trigger a test publish if we did `git tag test_1.23.45` for example

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,42 @@
+name: Publish Package
+
+on: push
+
+jobs:
+  build-n-publish:
```
rm-hull, line 6: maybe just -and- instead ?

thijstriemstra:
Btw, a new release would fix https://github.com/rm-hull/luma.core/discussions/226

.github/workflows/publish-to-pypi.yml
thijstriemstra: Can you give it a try..?

thijstriemstra: I would suggest only merging/testing the pypi test part for now by making a new release/tag and check if it's correctly published on test pypi. If it is, uncomment the release pypi in master and assume it will be fine when another release is made.

rm-hull (approved):

rm-hull:
https://github.com/rm-hull/luma.core/actions/runs/517360291

PR #234, opened 2021-01-28
ci: fix indentation

thijstriemstra:
refs #224


thijstriemstra:
It properly skipped the non-tag commit:

Screenshot from 2021-01-28 15-08-42

PR #237, opened 2021-03-07
Cleanup only used GPIO pins in serial interface

### LINKED ISSUE #137: Selectively cleanup RPi.GPIO pins
It's possible to pass an array of pins to `GPIO.cleanup(active_pins)`, e.g. `active_pins = [4, 5, 6]`. We should make sure `GPIO.cleanup()` only ever cleans up the active pins used by a luma library, not disabling other GPIO pins in the process.

See https://github.com/rm-hull/luma.core/blob/2.1.0/luma/core/interface/serial.py#L253

matemaciek:
When `luma.core.interface.serial.spi` is used alongside other modules using `RPi.GPIO`, they could collide if both are cleaning the same pins, therefore instead of cleaning all the pins let's clean only the ones we are using.

fixes #137 
fixes https://github.com/rm-hull/luma.oled/issues/325

### DIFF
--- a/luma/core/interface/serial.py
+++ b/luma/core/interface/serial.py
@@ -250,7 +250,9 @@ class bitbang(object):
         Clean up GPIO resources if managed.
         """
         if self._managed:
-            self._gpio.cleanup()
+            for pin in [self._SCLK, self._SDA, self._CE, self._DC, self._RST]:
+                if pin is not None:
+                    self._gpio.cleanup(pin)
 
 
 @lib.spidev
--- a/luma/core/interface/serial.py
+++ b/luma/core/interface/serial.py
@@ -361,6 +361,14 @@ class gpio_cs_spi(spi):
         if self._gpio_CS:
             self._gpio.output(self._gpio_CS, self._gpio.LOW if self._cs_high else self._gpio.HIGH)
 
+    def cleanup(self):
+        """
+        Close pin if it was set up.
+        """
+        if self._gpio_CS is not None:
+            self._gpio.cleanup(self._gpio_CS)
+        super(gpio_cs_spi, self).cleanup()
+
 
 class noop(object):
     """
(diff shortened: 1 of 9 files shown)


matemaciek:
This fixes https://github.com/rm-hull/luma.oled/issues/325. Tests fixes are coming (-:

rm-hull:
This a great improvement! Thank you. There are some build failures though, eg https://github.com/rm-hull/luma.core/pull/237/checks?check_run_id=2051211855#step:7:441

I can have a look at fixing if you can’t fathom what the issue is.

rm-hull:
Please also add yourself to the list of contributor in  CONTRIBUTING.rst

rm-hull (approved):

matemaciek:
I see that `cleanup` accepts also the list of ports, I'll change this PR to use it, as it will make both code and tests simpler. I'll also try to fix it for parallel connection.

luma/core/interface/serial.py
```diff
@@ -428,7 +426,7 @@ def output(self, pin, value):
 
         self._gpio.write(self._data)
 
-    def cleanup(self):
+    def cleanup(self, pin):
```
thijstriemstra, line 429: changing the public api should be avoided.

matemaciek: This is in following class:
```
class __FTDI_WRAPPER_GPIO:
    """
    Adapter for FTDI to RPI.GPIO. Not for direct public consumption
    """
```

thijstriemstra: ah yes, sorry for the noise.

rm-hull (approved):

rm-hull:
Released https://pypi.org/project/luma.core/2.3.1/

kudos to @thijstriemstra for the github pypi publish task - first time we've used it !

PR #238, opened 2021-03-21
tests: do not match exact OS errno strings

pinotree:
The actual string representation of an errno (like `ENOENT`) is an OS implementation detail, and there is no guarantee on it. Also, we are already checking that the exception is `FileNotFoundError`, which implies `ENOENT` already.

Hence, simplify the assertion checking: check that the actual filename that triggered the exception was the wrong one that was passed to `load_sprite_table()`.


dhrone (approved):
Good improvement to reduce OS coupling.

PR #246, opened 2021-09-23
serial.i2c: Only import smbus2 if no bus is provided

relvacode:
`smbus2` is always imported in `luma.core.interface.serial.i2c` even though it is not used if a bus is provided.

This means the i2c interface can't be used on Windows when given an alternate bus interface such as ftdi:

```
from luma.oled.device import ssd1309
from luma.core.interface.serial import ftdi_i2c

display = ssd1309(ftdi_i2c())
```

This PR tweaks things slightly to only import `smbus2` if no bus is provided.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -36,3 +36,4 @@ Contributors
 * Matthew Lovell (@mattblovell)
 * Tiago F. Pimentel (@antropoiese)
 * Maciej Sokolowski (@matemaciek)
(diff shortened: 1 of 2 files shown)


rm-hull (approved):
Makes sense! thanks for that
Could you add yourself to the list of contributors too please?

rm-hull (approved):

rm-hull:
Will push a new release out shortly
Thanks for taking the time to file this PR 🙇

PR #258, opened 2022-10-16
release: 2.4.0

thijstriemstra:


thijstriemstra:
I don't understand why github actions still wants to run the Python 3.6 process, even when it's not instructed to (any longer) 👎🏼 is there any cache or something that needs to be cleaned? Some of the checks in the list above have the label 'required' even though I don't see where this is specified...

thijstriemstra:
Build error is false broken link and can be ignored: https://github.com/rm-hull/luma.core/actions/runs/3261283157/jobs/5355843122#step:8:133

rm-hull (approved):

thijstriemstra:
hi @rm-hull can you tag this release so it gets published to pypi?

rm-hull:
Well, i tagged it, and pushed, the job eventually ran https://github.com/rm-hull/luma.core/actions/runs/3276977398, but didnt actually publish

rm-hull:
https://pypi.org/project/luma.core/2.4.0/ released manually

PR #267, opened 2023-09-01
update .readthedocs.yml to v2

thijstriemstra:


rm-hull (dismissed):

rm-hull (dismissed):

thijstriemstra:
Looks like the doc build is failing on Python 3.7 because it's using an old version of Sphinx (because python 3.7 support was dropped in sphinx v6.0 so no way to upgrade it either. The actual string it's failing on is not coming from luma.core but the intersphinx for Pillow, which likely was fixed in Sphinx at some point.

.github/workflows/main.yml
```diff
@@ -39,6 +40,7 @@ jobs:
           python_env=$(echo $TOX_ENV | sed -e s/-dev$//)
           python -m tox -e ${python_env}
       - name: QA
+        if: ${{ steps.setuppy.outputs.python-version > 7 }}
```
rm-hull, line 43: 🤫

rm-hull (approved):

Issue #268, opened 2023-12-16
Raspberry Pi 5 + SPI Display 


brickbots:
Hello!

I'm working to port some working code to bookworm on a Raspberry Pi 5.  This code works great on bookworm/RPI4, but with the RPI5 it's not able to initialize the SPI interface from the luma.core.interface.serial module.

Perhaps this is due to the new hardware controller on the 5?  

#### Type of Raspberry Pi
Raspberry Pi 5

#### Linux Kernel version
Bookworm
Linux pifinder 6.1.0-rpi7-rpi-v8 #1 SMP PREEMPT Debian 1:6.1.63-1+rpt1 (2023-11-24) aarch64 GNU/Linux

#### Expected behaviour
I'm trying to initialize a waveshare 1.5" RGB OLED display based on the ssd1351 controller.  Communication is via SPI

```python
from luma.core.interface.serial import spi

# init display  (SPI hardware)
serial = spi(device=0, port=0)
```

#### Actual behaviour
Crashes with the error below

 serial = spi(device=0, port=0)
             ^^^^^^^^^^^^^^^^^^^^^
  File "...python/venv/lib/python3.11/site-packages/luma/core/interface/serial.py", line 302, in __init__
    bitbang.__init__(self, gpio, transfer_size, reset_hold_time, reset_release_time, DC=gpio_DC, RST=gpio_RST)
  File "...python/venv/lib/python3.11/site-packages/luma/core/interface/serial.py", line 190, in __init__
    self._DC = self._configure(kwargs.get("DC"))
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...python/venv/lib/python3.11/site-packages/luma/core/interface/serial.py", line 203, in _configure
    self._gpio.setup(pin, self._gpio.OUT)
RuntimeError: Cannot determine SOC peripheral base address

thijstriemstra:
From https://forums.raspberrypi.com/viewtopic.php?t=361218:

> RPi.GPIO, wiringPI, and pigpio do not work on the Pi5 which has new hardware for the GPIO.

Yikes..

brickbots:
Thank you!  Yes I've tried and no it does not work... I've played a little bit with GPIOD and it's related python bindings, which seems to be the method that people are coalescing around.  It theoretically works for RPI 3/4/5 and is fairly quick.... but I've not seen anything about SPI or I2C comms using that framework, just raw GPIO on/off.  

[...]

PR #269, opened 2024-01-04
Study of a possible implementation of gpiozero to support RPI5

outdoorbits:
This is just a study of a possible implementation of gpiozero to support the Raspberry Pi 5 in the future. Up to now I didn't step deep enough into your code to really understand it. So I couldn't test it yet. But I think this could be an easy way to migrate from RPi.GPIO to gpiozero. The advantage would be to have one library (gpiozero) which is compatible to both, Raspberry Pi 4 (and older) and Raspberry Pi 5 (and hopefully others in future). 

refs #268


thijstriemstra:
thanks! I saw another project (moonraker) that switched to https://pypi.org/project/python-periphery, maybe an option as well..

outdoorbits:
But the last commit of python-periphery is much older than the RPi 5, so it won't be able to support its new hardware.

thijstriemstra:
> But the last commit of python-periphery is much older than the RPi 5, so it won't be able to support its new hardware.

Not sure, as long as it supports the linux kernel gpio interface (around for years), it should be fine since that is what the rpi5 uses afaik.

outdoorbits:
I don't think python-periphery works. I tested inputs:

```
from gpiozero import Button
from signal import pause

def pressed():
        print('+')

def released():
        print('-')


button = Button(26,pull_up=False)

button.when_pressed = pressed
button.when_released = released

pause()
```

works find while

```
from periphery import GPIO
import time

# Open GPIO /dev/gpiochip0 line 10 with input direction
gpio_in = GPIO("/dev/gpiochip0", 26, "in")
# Open GPIO /dev/gpiochip0 line 12 with output direction
gpio_out = GPIO("/dev/gpiochip0", 12, "out")

while True:
        print(gpio_in.read())
        time.sleep(1)

gpio_out.write(not value)

gpio_in.close()
gpio_out.close()
```
didn't work.

PR #274, opened 2024-01-21
tests: update images for Pillow 10.1

thijstriemstra:
I updated the test images but the results look quite different, because Pillow 10.1 is using a different and [new builtin font](https://pillow.readthedocs.io/en/stable/releasenotes/10.1.0.html#size-and-font-size-arguments-when-using-default-font). I'll see if the old one can be used instead.


thijstriemstra:
> I'll see if the old one can be used instead.

I added the old font and tests work again as expected.

rm-hull (approved):

PR #277, opened 2024-05-18
Raspberry Pi 5 Support

RampantDespair:
This is a fork of a fork (https://github.com/outdoorbits/luma.core) in hopes to add RPi 5 support to luma.core


RampantDespair:
Added `gpiozero` to setup requirements because the tests seem to be failing primarily is relation to `No module named 'gpiozero'`

PR #284, opened 2024-10-26
Make GPIO/SPI/FTDI optional and switch to rpi-lgpio

### LINKED ISSUE #270: RPI.GPIO should be optional

thijstriemstra:
fixes #270 

to install optional dependencies:

- GPIO : `pip install luma.core[gpio]`
- SPI: `pip install luma.core[spi]`
- FTDI: `pip install luma.core[ftdi]`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,6 +33,15 @@ jobs:
         run: python -c "import sys; print(sys.version)"
       - name: Install Python packages
         run: pip install --upgrade setuptools pip wheel tox coveralls
+      - name: Install system dependencies
+        run: sudo apt-get install swig python3-setuptools python3-dev
+      - name: Install lg library
+        run: |
+          wget http://abyz.me.uk/lg/lg.zip
+          unzip lg.zip
+          cd lg
+          make
+          sudo make install
       - name: Run tests
         env:
           TOX_ENV: py3${{ matrix.python-minor-version }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,6 +37,7 @@ jobs:
         run: sudo apt-get install swig python3-setuptools python3-dev
       - name: Install lg library
         run: |
+          cd $RUNNER_TEMP
           wget http://abyz.me.uk/lg/lg.zip
           unzip lg.zip
           cd lg
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-minor-version: [8, 9, 10, 11, 12, 13, 14-dev]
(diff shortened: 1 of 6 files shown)


thijstriemstra:
@Pako2 @rm-hull  I finally got my hands on a RPI5 and going through the tests to see what breaks and what doesn't and fortunately everything still works when switching to rpi-lgpio, a (supposedly) drop-in replacement for rpi.gpio. I haven't actually tested it with a real display but tests run and pass on an RPI5 for now. 

I also started a RPI5 PR for luma.oled: https://github.com/rm-hull/luma.oled/pull/394

rm-hull (approved):
Lgtm

Pako2:
I assume that there are a number of existing users (luma libraries) whose display stopped working after the system upgrade, because the sysfs GPIO interface (and therefore the RPi.GPIO based on it) was no longer supported.
For these users, currently switching to **rpi-gpio** (which replaces RPi.GPIO) is probably the only meaningful way to maintain functionality without the need for any complex modifications.
Unfortunately, this only applies to Raspberry Pi SBC users. However, there are also users of SBCs from other manufacturers, such as Allwinner. For those, there was an alternative OPi.GPIO library that was compatible with RPi.GPIO. This allowed luma libraries to be used to control OLED displays. I myself am one of these users and after upgrading LibreELEC/Kodi my display stopped working. For Allwinner (and probably for no other type of SBC) there is nothing like **rpi-gpio**. I think there is no point in working on creating something like _opi-gpio_. As I suggested earlier in [#278](https://github.com/rm-hull/luma.core/pull/278), I consider it far more promising to completely skip this intermediate link (RPi.GPIO/**rpi-lgpio**) and use **lgpio** directly. This makes much more sense and has one huge advantage: it allows you to work with luma libraries also on SBCs from other manufacturers (for example Allwinner, Amlogic, Rockchip ...). Specifically for Allwinner, I already tried it and it works very well...
**So I have a question: does anyone deal with this idea?**
Unfortunately, no one responds to my posts in this sense in [#278](https://github.com/rm-hull/luma.core/pull/278).

peteS-UK:
I wondered if there were plans to merge this one?  I use luma.core in a couple of projects for screen display with squeezebox and they can't work with rpi5 currently.  Thanks for all the work :-)

PR #290, opened 2025-11-06
cmdline.create_parser(): Allow pre-created parser

mdevaev:
Additionally, it is possible to disable the automatic initialization of autocomplete.

Context: I need this to simply add luma arguments to my pre-created parser [like here](https://github.com/pikvm/kvmd/blob/master/kvmd/apps/otgconf/__init__.py#L288) that initializes the system options. The proposed PR does not break backward compatibility.

### DIFF
--- a/luma/core/cmdline.py
+++ b/luma/core/cmdline.py
@@ -268,12 +268,24 @@ def create_device(args, display_types=None):
     return device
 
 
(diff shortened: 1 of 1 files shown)


mdevaev:
Merci :heart:

PR #292, opened 2025-12-07
Added support for AiP31068 and PCA9633

CutterXYZ:
Support for AiP31068.
This is a standard I2C controller for HD44780 LCD displays.

Spec sheet:
https://www.orientdisplay.com/wp-content/uploads/2022/08/AIP31068L.pdf

Support for PCA9633.
This is a RGB backlight controller over a dedicated I2C port.

Spec sheet: 
https://www.nxp.com/docs/en/data-sheet/PCA9633.pdf

This has been tested with the Waveshare LCD1602 display:
https://www.waveshare.com/wiki/LCD1602_RGB_Module


CutterXYZ:
Thanks, I'm OK with it.

PR #301, opened 2026-02-01
ci: cache lg build

thijstriemstra:
cache the download and build of lg to greatly ease the strain on someone else's server.


rm-hull (dismissed):

rm-hull (approved):
👍

